Issue #20, opened 2015-09-09
Clean up (Cabal) package


jpvillaisaza:

jpvillaisaza:
Done, see https://github.com/stackbuilders/hapistrano/commit/c6fb4becfcd39bb2751b32e8cef9c505779c0000.

PR #22, opened 2015-09-09
Clean up (Cabal) package

jpvillaisaza:
See #20.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -7,7 +7,7 @@ import System.Environment.Compat (lookupEnv)
 import System.IO (hPutStrLn, stderr)
 import System.Exit (exitFailure)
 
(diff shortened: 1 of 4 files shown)


jpvillaisaza:
@jsl What do you think?

The executable is now in the app directory and has the library as a build dependency.

One thing I don't like is that I had to expose System.Hapistrano.Types because the test suite uses it. Instead, we could add the src directory to the hs-source-dirs of the test suite or re-export System.Hapistrano.Types in System.Hapistrano to keep only one exposed module. I think I like the latter, re-exporting.

jsl:
LGTM!

jpvillaisaza:
@jsl I'll merge once the checks are complete. I re-exported `ReleaseFormat(..)` in `System.Hapistrano`, that's the only type used in the tests right now.

PR #24, opened 2015-09-09
Upgrade Travis CI to container-based infrastructure

jpvillaisaza:


jsl:
LGTM, but looks like we have to resolve some issues with build failures. Feel free to merge when it's ready. :)

jpvillaisaza:
@jsl Yes, thank you. Do you know if the tests were disabled in Travis for a particular reason? There was no `cabal test` there.

jpvillaisaza:
@jsl I think Travis should work now. The tests are still not there. They're failing, but only on Linux, not Mac. I opened #25 to fix that.

Issue #25, opened 2015-09-09
Tests fail on Linux


jpvillaisaza:
The tests fail on Linux (they work on Mac).

jpvillaisaza:
The way we're running some commands is `readProcessWithExitCode`, but we should use something like `readCreateProcessWithExitCode` to have more control of the command. To create the process, we can use either `proc` or `shell`. Right now, we're using the equivalent of `proc` and simply taking the head of a list to be the raw command and the rest of the list to be the argument list. That causes commands like `git commit` to fail. Since we're simply passing a complete command, we can use `shell` instead.

Additionally, `runCommand` is a function in both System.Hapistrano and System.HapistranoSpec. Perhaps we can keep only one of them.

There are some pieces missing with `git` commands to get the tests passing as well, but I don't have a solution for those yet.

jpvillaisaza:
See #26 for the fix.

jpvillaisaza:
Fixed. See 4150310, baccedd, and ada5961.

PR #26, opened 2015-09-10
[WIP] Fix tests on Linux

jpvillaisaza:
See #25.

### DIFF
--- a/spec/System/HapistranoSpec.hs
+++ b/spec/System/HapistranoSpec.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE CPP #-}
+
 module System.HapistranoSpec (spec) where
 
 import Test.Hspec (it, describe, shouldBe, Spec)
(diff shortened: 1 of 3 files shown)


spec/System/HapistranoSpec.hs
```diff
@@ -40,7 +40,7 @@ genSourceRepo path = do
         , "echo testing > " ++ joinPath [clonePath, "README"]
         , "cd " ++ clonePath ++ " && git add -A"
         , "cd " ++ clonePath ++ " && git commit -m\"First commit\""
-        , "cd " ++ clonePath ++ " && git push"
+        , "cd " ++ clonePath ++ " && git push origin master"
```
jpvillaisaza: @jsl Is `"origin master"` correct? `git push` fails in my computer unless I specify what default behavior I want. I don't know if that's the case with the Git installed in Travis.

jpvillaisaza:
`readCreateProcessWithExitCode` is very new... It doesn't work for GHC < 7.10 I think, but it fixed the problem for GHC 7.10.2. I'll take a look at how to fix the other builds tomorrow.

spec/System/HapistranoSpec.hs
jsl: Hm, maybe defaults are different across machines. Since this is just a test repo that we blow away anyway, seems like it should be fine to be explicit.

jpvillaisaza:
@jsl The tests are now passing for GHC 7.10.2 on both Linux and Mac, but failing for GHC 7.6.3 and 7.8.4. I disabled those tests for now, maybe I'll try fixing them later at home so that I don't have to keep waiting on Travis.

jpvillaisaza:
I'm closing this for now. I'll get back to it this week with Linux and try again. The only problem is making the tests pass for GHC < 7.10.

PR #29, opened 2015-09-23
Fix tests on Linux

jpvillaisaza:
See #25.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ install:
     fi
   - travis_retry cabal update -v
   - sed -i 's/^jobs:/-- jobs:/' ${HOME}/.cabal/config
-  - cabal install --only-dependencies --dry -v > installplan.txt
+  - cabal install --only-dependencies --enable-tests --dry -v > installplan.txt
   - sed -i -e '1,/^Resolving /d' installplan.txt; cat installplan.txt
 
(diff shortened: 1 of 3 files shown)


jpvillaisaza:
@jsl @juanpaucar Can you take a look?

The use of `System.Process.readProcessWithExitCode` fails for built-in commands such as `cd` on Linux. The application actually works because it calls `readProcessWithExitCode` with `ssh`, but the tests fail. The new `System.Process.readCreateProcessWithExitCode` function would fix the issue, but only for GHC 7.10.2. The solution right now is to change `System.HapistranoSpec.runCommand` to manually create the process with `System.Process.shell` (not `proc`), and `System.Hapistrano.runCommand` to call a new `execShellCommand` only when there's no SSH server.

So, there's no code specific to any operating system right now and the tests are passing on both Mac and Linux. In order to fix the Travis build, I had to add a user.email, user.name, and set the push.default to simple (the push.default was only a problem with Travis, I was only getting a warning on my Linux computer).

The solution seems to work for now. It could be refactored so that we don't have two separate functions running commands (the ones in the tests are in IO, the ones in Hapistrano are in the Hapistrano type).

juanpaucar:
@jpvillaisaza LGTM Good Job :+1:

jsl:
LGTM, thanks!

Issue #31, opened 2015-09-24
Test suite run failure from tarball


snoyberg:
I see error messages such as the following:

```
Test suite hapistrano-test: RUNNING...
Deploy failed with (status, message): (128,"fatal: ambiguous argument 'master': unknown revision or path not in the working tree.\nUse '--' to separate paths from revisions, like this:\n'git <command> [<revision>...] -- [<file>...]'")
Deploy failed with (status, message): (128,"fatal: ambiguous argument 'master': unknown revision or path not in the working tree.\nUse '--' to separate paths from revisions, like this:\n'git <command> [<revision>...] -- [<file>...]'")
```

jpvillaisaza:
Closing now, the problem should be fixed by ca5c7a2.

PR #32, opened 2015-09-24
Fix Git configuration for tests

jpvillaisaza:
See #31.


jpvillaisaza:
@juanpaucar Can you take a look? Do you think this change makes sense?

juanpaucar:
Was the missing git configuration what made the tests fail in the stackage travis environment?

jpvillaisaza:
@juanpaucar Yes, the Git configuration for push.default, user.email, and user.name was in our .travis.yml, so it worked only here.

juanpaucar:
LGTM :shipit:

Issue #33, opened 2015-09-24
Handle command line options with Turtle or GetOpt


jpvillaisaza:
We can handle command line options with turtle or simply System.Console.GetOpt. In either case, it'll be easier to add new options and we can appropriate help and usage messages. Additionally, the application should print a more useful message if one of the required environment variables is not set instead of simply relying on the exception raised by the getEnv function.

PR #34, opened 2015-09-24
Handle command line options with GetOpt

jpvillaisaza:
See #33.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -1,13 +1,20 @@
-module Main where
+{-# LANGUAGE RecordWildCards #-}
+
+module Main (main) where
 
 import qualified System.Hapistrano as Hap
 import Control.Monad (void)
-import System.Environment (getArgs, getEnv)
+import System.Environment (getEnv)
 import System.Environment.Compat (lookupEnv)
-import System.IO (hPutStrLn, stderr)
-import System.Exit (exitFailure)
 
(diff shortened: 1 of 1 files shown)


jpvillaisaza:
@jsl @juanpaucar Can you take a look? This adds handling of command line options with System.Console.GetOpt (I used hapistrano as a test case for a blog post that I'm working on right now) and handles the case in which an environment variable is missing (instead of throwing an exception).

juanpaucar:
I'm not sure about the use of `fromJust` I know that we are never going to hit an error because of [these lines](https://github.com/stackbuilders/hapistrano/pull/34/files#diff-63e7d2cbf23880a9c902c91fa34f6043R50) and using other ways of dealing with the `Nothing` would be unnecessary but it kinda looks weird.

But the overall LGTM

jpvillaisaza:
@juanpaucar The use of `fromJust` was safe there, but not necessary. I force-pushed some changes to use `maybe` instead. It's prettier and simpler now.

juanpaucar:
Indeed LGTM :shipit:

jsl:
I just noticed this PR - sorry for the delay. @jpvillaisaza I'm fine with merging this unless you see a reason not to.

jpvillaisaza:
@jsl Done. Thanks!

Issue #35, opened 2016-02-24
Have ability to deploy by different ssh port


lpaulmp:
I was not allowed to deploy using a different ssh port. e.g

```
export HOST="user@localhost:2222"
```

mrkkrp:
Resolved via #39.

PR #38, opened 2016-03-04
Parse port when is in the HOST

CristhianMotoche:
I added a function that checks if the port is in the HOST, e. g. `user@machine:port`. If so, then it executes `ssh user@machine -p port`, otherwise it simple executes `ssh user@machine`.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -8,6 +8,7 @@ import System.Environment.Compat (lookupEnv)
 
 import System.Hapistrano (ReleaseFormat(..))
 
+import qualified Control.Applicative as Applicative
 import qualified Control.Monad as Monad
 import qualified Data.Maybe as Maybe
 import qualified System.Console.GetOpt as GetOpt
@@ -51,7 +52,7 @@ configFromEnv = do
   repository <- maybe (Exit.die (noEnv "REPOSITORY")) return maybeRepository
   revision <- maybe (Exit.die (noEnv "REVISION")) return maybeRevision
 
-  host           <- fmap parseHostWithPort $ lookupEnv "HOST"
+  host           <- fmap parseHostWithPort Applicative.<$> lookupEnv "HOST"
   buildScript    <- lookupEnv "BUILD_SCRIPT"
   restartCommand <- lookupEnv "RESTART_COMMAND"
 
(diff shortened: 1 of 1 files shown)


jsl:
Note that this is for the issue specified in #35.

app/Main.hs
```diff
@@ -145,3 +145,17 @@ main = do
     Just HapRollback -> rollback hapConfiguration
 
     Nothing -> hapHelpAction Nothing
+
+splitInColon :: String -> (String, String)
```
jsl: Thanks @CristhianMotoche for making this change. The only thing I'm not crazy about is the conditional check to see if the host is `null`. What do you think about giving this the following return type, against which we could pattern match?

```haskell
splitInColon :: String -> (String, Maybe String)
```

At least that's what initially came to mind for me. @jpvillaisaza do you have any other ideas?

jpvillaisaza: @jsl @CristhianMotoche We could use something more general, like [`breakOn`](https://www.stackage.org/haddock/lts-5.5/extra-1.4.3/Data-List-Extra.html#v:breakOn) from extra:

```
breakOn :: Eq a => [a] -> [a] -> ([a], [a])
```

(We don't have to include the `extra` package, using `break` would be enough if that's all we need.)

I think using a list (and an empty list for no port) is fine here (we could turn it into a `Maybe` if we want to, but I don't think it makes a big difference).

We could use a `case` instead of a `let` and an `if`. Something like this:

```
case breakOn ':' server of
  (host, []) -> host
  (host, (_:port)) -> unwords [host, "-p", port]
```

The type signature I don't like is the one for `parseHostWithPort`. I think it would be better to make it

```
parseHostWithPort :: String -> String
```
 and use a second `fmap` (or `<$>`) above or anywhere else where a `Maybe` could make sense.

jsl: @jpvillaisaza :+1:

CristhianMotoche:
I was thinking about this PR, and I thought that could be a better to read an env variable called `PORT`, if that variable is not defined, we can use the default `22` port. What do you think?

jsl:
@CristhianMotoche yeah that seems a bit more consistent with the way that we do things currently.

lpaulmp:
@CristhianMotoche +1 to read an env var.

CristhianMotoche:
I think, I'll close this one. Because of #39.

PR #39, opened 2016-04-14
Parse env variable PORT

CristhianMotoche:
I lookup the environmental variable `PORT` to deploy the application to a different ssh port if it is present. If it is not, then we take the default port. So, the variable `PORT` is optional. To solve #35


jpvillaisaza:
@CristhianMotoche Does this pull request replace #38?

CristhianMotoche:
Yes @jpvillaisaza. I'll close the other PR, if we think that this approach is better. Otherwise, I'll try to improve the other PR.

Issue #41, opened 2016-10-19
Support comments and empty lines


juanpaucar:
Currently hapistrano will fail if there are comments in the building files. The same will happen with empty lines.

This should be a enhancement since sometimes it's better to have comments and empty lines that make easier the reading. This is related to how we pass the lines arguments to the `buildRelease` function 

``` Haskell
buildRelease :: Release  -- ^ The Release to build
             -> [String] -- ^ Commands to be run. List intercalated
                         -- with "&&" so that failure aborts the
                         -- sequence.
             -> Hapistrano ()
buildRelease rel commands = do
  conf <- ask
  let cdCmd = "cd " ++ releasePath conf rel
  void $ runCommand (host conf) $ intercalate " && " $ cdCmd : commands

```

mrkkrp:
Resolved via #44.

PR #44, opened 2016-10-20
Support comments and empty lines, solves #41

juanpaucar:
@jpvillaisaza @sestrella could you take a look please?

### DIFF
--- a/src/System/Hapistrano.hs
+++ b/src/System/Hapistrano.hs
@@ -8,6 +8,7 @@ module System.Hapistrano
 
        , activateRelease
        , currentPath
+       , cleanBuildScript
        , defaultSuccessHandler
        , defaultErrorHandler
        , directoryExists
@@ -36,8 +37,8 @@ import Control.Monad.Trans.Either ( left
                                   , right
                                   , eitherT )
 
-import Data.Char (isNumber)
-import Data.List (intercalate, sortBy, isInfixOf)
+import Data.Char (isNumber, isSpace)
+import Data.List (intercalate, sortBy, isInfixOf, dropWhileEnd)
 import Data.Time (getCurrentTime)
 import Data.Time.Format (formatTime)
 import Data.Time.Locale.Compat (defaultTimeLocale)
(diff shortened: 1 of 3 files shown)


src/System/Hapistrano.hs
```diff
@@ -330,6 +331,13 @@ restartServerCommand = do
     Nothing -> return "No command given for restart action."
     Just cmd -> runCommand (host conf) cmd
 
+cleanBuildScript :: [String] -> [String]
+cleanBuildScript allScriptLines = filter (not . isCommentOrEmpty) allScriptLines
+  where
+    isCommentOrEmpty line = isEmpty line || isComment line
```
jsl: Does this work if there is leading space on the function?

We may want to talk about what the code is doing here. There may be a better approach, perhaps using megaparsec's code to ignore comments, or by implementing this in a different way altogether.

juanpaucar: Indeed,  It trims these chars `\t, \n, \r, \f, \v` as defined by is the function [isSpace](https://hackage.haskell.org/package/base-4.9.0.0/docs/Data-Char.html#v:isSpace).
I think using a parsec might be an overkill since hapistrano is pretty lightweight in dependencies.

\cc: @jpvillaisaza any thoughts on this?

jpvillaisaza: @jsl @juanpaucar What should happen if there's a leading space? I don't fully understand the original question.

As far as adding megaparsec as a dependency, it wouldn't be a big change in number of dependencies. Just about two additional things to install (at least with a very recent set of packages).

jsl:
Honestly, I can't recall why having comments or spaces breaks hapistrano in the first place. Instead of creating something that makes a bash script into something that can be executed by hapistrano, should we just scp the script to the server and eval it there using bash directly?

jsl:
Actually, in most of our use cases, I think this script is part of the code repo, so it's already on the server. Perhaps we just change so that it can execute a script on the remote server, rather than sending the commands that we find in the file?

juanpaucar:
What about something more like capistrano or [shipit](https://github.com/shipitjs/shipit) where we write task that should be executed remotely. Although, this probably will be a big change for hapistrano

jpvillaisaza:
I think it breaks it because it takes each line, intercalates `&&`, and then executes that as one command.

juanpaucar:
@jpvillaisaza Exactly that's why I only wanted to filter out comment lines and empty lines. Also trimming them in the process

mrkkrp:
Merging this as the PR is getting rather old. I'll probably refactor it and use Megaparsec though.

jsl:
@mrkkrp I think that this wasn't merged because we were worried about [this edge case](https://github.com/stackbuilders/hapistrano/pull/44#issuecomment-255428948). Could you check it please? If it doesn't have tests to cover things like this we may want to add them. Thanks!

mrkkrp:
I'm going to add support for comment and tests, don't worry. I'm just merging anything so we don't have hanging work before I start refactoring, etc.

Issue #43, opened 2016-10-19
Add command line arguments


jpvillaisaza:
A part of #28 is about adding command line arguments such as `--help` to at least display some usage information. I chose cmdargs for a project I added command line arguments recently, but a similar one would work as well. Basic options (help, version, ...) are good enough to start, but there are a lot of nice ones we can add.

Some ideas after a quick look at Capistrano:
- `--dry-run` to do a dry run
- `--print-config-variables` to display the configuration before deploying

mrkkrp:
I would vote for `optparse-applicative` as well, it's hard to find anything better than this.

jpvillaisaza:
@CristhianMotoche One of the things I like about cmdargs is that is has almost no dependencies (base, filepath, process, template-haskell, and transformers, which are all boot libraries). But this is not a very important advantage over optparse-applicative (which has as non-boot library dependencies ansi-wl-pprint, transformers-compat, and, in some cases, semigroups).

So, both are fine as far as I'm concerned. I didn't know about bash completion with optparse-applicative, which is interesting and probably worth it (even though bash completion might not be that important for this program).

I would vote for cmdargs, which is the one I suggested, but you can go ahead with optparse-applicative following @mrkkrp's vote (and I'm guessing more people will vote for that one since it's the one we use for other programs such as dotenv).

@CristhianMotoche Would you like to start working on this and open a pull request?

CristhianMotoche:
Hey, sorry for the late response, but sure I'd like to work on this.

jpvillaisaza:
@CristhianMotoche Awesome, go ahead.

jpvillaisaza:
It seems like we're using System.Console.GetOpt, but it's just a usage message without much information, so moving to something else makes more sense.

jsantos17:
+1 optparse-applicative. @CristhianMotoche, let me know if you have any
issues :).

mrkkrp:
Resolved via #46.

PR #46, opened 2016-10-21
Add optparse-applicative for parsing args

CristhianMotoche:
Sorry, for taking too long. It's the first time I use this library.

@jpvillaisaza Could you tell me please if I'm in the right way? Thanks in advance.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -12,7 +12,7 @@ import qualified Control.Monad as Monad
 import qualified System.Exit.Compat as Exit
 
 import Data.Monoid((<>))
-import Options.Applicative(ParserPrefs(..))
+import Options.Applicative(ParserPrefs(..), (<|>))
 import qualified Options.Applicative as Opt
 
 import Paths_hapistrano (version)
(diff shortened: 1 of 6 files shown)


jpvillaisaza:
@CristhianMotoche I haven't reviewed in detail yet, but I tried your branch and it the messages are starting to look much better than what we have right now.

CristhianMotoche:
@jpvillaisaza The current application only has 2 commands (`deploy` and `rollback`) and `help` flag. Should we open different PRs for the flags that you mentioned in #43? or should we add those in this PR?
I'd prefer to add them in different PRs. Well, let me know what you think about it.

jpvillaisaza:
@CristhianMotoche Yes, this can be just for deploy, help, and rollback and new ones in separate issues/pull requests.

jsl:
Thanks @CristhianMotoche and @jpvillaisaza - I'm really glad to see this library improving.

hapistrano.cabal
```diff
@@ -51,6 +51,7 @@ executable hap
   build-depends:       base
                      , hapistrano
                      , base-compat
+                     , optparse-applicative >= 0.13 && < 0.14
```
jpvillaisaza: @CristhianMotoche Does this work with optparse-applicative 0.12 (the one in the most recent LTS snapshot)? We could also test with 0.11 (the one in LTS 3.5, which is the one specified in the stack.yaml file) (and we can set the resolver to something more recent). If they all work, we can change this constraint to something like >= 0.11 && < 0.14 and remove the extra-deps section from the stack.yaml file.

Other than that, this looks good. Let me know if it's ready and we can accept this. I haven't used optparse-applicative, so I'll check the documentation and review again if it's ready.

mrkkrp:
@CristhianMotoche @jpvillaisaza Hey guys, don't be too fast to merge this, I want to take a look tomorrow and leave some comments :-)

jpvillaisaza:
@mrkkrp No problem, we can wait (and you have experience with optparse-applicative so you're review can be more helpful here). And I'd like the other thing I mentioned to be ready before merging so that you can print the help message and version without having to set environment variables.

app/Main.hs
```diff
@@ -66,82 +68,25 @@ configFromEnv = do
...
+          Opt.<|> Opt.flag' printVersion (Opt.long "version" <> Opt.short 'v' <> Opt.help "Diplay the version of Hapistrano")
+      commands hapConfig =
+        Opt.command "deploy"
+          (Opt.info (Opt.pure $ deploy hapConfig) (Opt.progDesc "Deploys the current release with the configure options"))
+        <> Opt.command "rollback"
+          (Opt.info (Opt.pure $ rollback hapConfig) (Opt.progDesc "Rolls back to the previous release"))
+
+prefs :: ParserPrefs
+prefs =
+  Opt.defaultPrefs { prefShowHelpOnEmpty = True }
```
CristhianMotoche: I agree with you. I don't like to set the env variables to then see the help information. I'll see what can I do about it. Thanks @jpvillaisaza

CristhianMotoche:
Of course @mrkkrp 
I'd like to learn from your comments. :-)

CristhianMotoche:
I'm not sure if my last commit is fine. However, we can always do `git reset --hard HEAD~1 && git push camm-origin -f` to revert it. Just let me know it.

hapistrano.cabal
CristhianMotoche: I'm going to tested with `optparse-applicative 0.12`.

However, I think the `Opt.defaultPrefs { prefShowHelpOnEmpty = True }`  isn't available in `optparse-applicative 0.11`. That is why I didn't use that version.

Take a look: [optparse-applicative](http://hackage.haskell.org/package/optparse-applicative-0.11.0.2/docs/Options-Applicative-Extra.html#t:ParserPrefs)

mrkkrp:
Two things here.

1. Optional, just an opinion. I see the code is already suffering from this, but would be nice to use non-qualified import from `Options.Applicative`, because `Opts` prefix before every and each function really makes it hard to read. I see someone even added `Control.Monad` as `Monad`, which I find outrageous, one should know standard vocabulary, so `Monad` prefix is certainly overkill.

2. Now about actual code. You don't need `customExecParser` at all, and you don't need `defaultPrefs`. You have `opts` of type `Parser ()` and you want to:

   ```haskell
   main :: IO ()
   main =
     join $ Opt.execParser (Opt.info (Opt.helper <*> opts) desc)
       where
         opts =
           Opt.subparser commands
           <|> Opt.flag' printVersion
           (Opt.long "version" <> Opt.short 'v' <> Opt.help "Diplay the version of Hapistrano")
         commands =
           addCommand deployCommand
           <> addCommand rollbackCommand

         desc = Opt.fullDesc <> Opt.progDesc "…" <> Opt.header "…"
   ```

   The `helper` thing will add `--help` option to your parser that displays help. When user gives invalid options, the help is also displayed, IOW the behaviour is OK out-of-the-box, and you can use `optparse-applicative-0.11` I guess. To make things prettier, you should add program description in `desc` as well.

CristhianMotoche:
Thanks for your comments @mrkkrp :+1:

CristhianMotoche:
Sorry, I didn't have enought time to follow up this PR. However, today I had some free time and I pushed something new. Tell me what you think about it.

mrkkrp:
I think this has been hanging here for too long. I'm merging this for now and closing #43. I'll probably refactor this though.

CristhianMotoche:
@mrkkrp Cool, just let me know what can I do to improve this functionality. :+1:

PR #48, opened 2016-11-14
Add version constraints to dependencies

### LINKED ISSUE #42: Add version constraints
Right now, the only version constraints in the Cabal file are the ones for the base package. We should add constraints (lower and upper bounds) to most dependencies.

nebtrx:


nebtrx:
@jpvillaisaza have a look at this one and provide feedback

hapistrano.cabal
```diff
@@ -35,22 +35,22 @@ library
...
-                     , time-locale-compat
-                     , time
-                     , transformers
+                     , either >= 4.0 && < 4.6
+                     , filepath >= 1.3 && < 1.5
+                     , mtl >= 2.0 && < 3.0
+                     , process >= 1.2 && < 1.5
+                     , time-locale-compat >= 0.1 && < 1.0
+                     , time >= 1.4.1 && < 2.0
+                     , transformers > 0.3 && < 1.0
```
jpvillaisaza: We don't need version constraints for boot libraries (filepath, process, time, and transformers). I don't think we need a particular version for any of those, so the version constraints for base should be enough. The ones for either, mtl, and time-locale-compat are fine.

hapistrano.cabal
```diff
@@ -35,22 +35,22 @@ library
...
   ghc-options:         -Wall
   default-language:    Haskell2010
 
 executable hap
   hs-source-dirs:      app
   main-is:             Main.hs
-  build-depends:       base
+  build-depends:       base >= 4.5 && < 4.10
```
jpvillaisaza: This one for base can be removed as well. Since the library has specific version constraints, that should be enough.

hapistrano.cabal
```diff
@@ -59,15 +59,15 @@ test-suite hapistrano-test
...
-                     , mtl
-                     , process
-                     , temporary
+                     , directory >= 1.2.1 && < 1.3
+                     , either >= 4.0 && < 4.6
+                     , filepath >= 1.3 && < 1.5
+                     , hspec > 2.0 && < 2.5
+                     , mtl >= 2.0 && < 3.0
+                     , process >= 1.2 && < 1.5
+                     , temporary >= 1.2 && < 1.5
```
jpvillaisaza: Same here. We don't need version constraints for base, directory, filepath, and process.

nebtrx:
@jpvillaisaza changes applied

jpvillaisaza:
@nebtrx Thanks!

PR #50, opened 2016-11-16
Integration with shake

sestrella:


sestrella:
I'm closing this PR since we need to have more debate about this approach

PR #52, opened 2017-02-03
Refactor/rewrite all the code

mrkkrp:
The change prepares solid ground for further improvements. This also fixes at least two bugs:

* Proper support for empty lines and comments in build scripts.

* Previously switching to new releases was not atomic despite all the symlink business. This is because it deleted the current link before creating the new one, but it should have created a new link and then use `mv` to replace the old one atomically. The code has been changed to work properly now.

Tests are comping soon, I'll need to figure out what to port and what to add from scratch, so it's still WIP.


mrkkrp:
@sestrella Perhaps you could take a look and tell me your thoughts so far?

src/System/Hapistrano/Core.hs
```diff
@@ -0,0 +1,98 @@
...
+-- | Run the given sequence of command. Whether to use SSH or not is
+-- determined from settings contained in the 'Hapistrano' monad
+-- configuration. Commands that return non-zero exit codes will result in
+-- short-cutting of execution.
+
+exec :: forall a. Command a => a -> Hapistrano (Result a)
+exec typedCmd = do
+  Config {..} <- ask
+  let (prog, args) =
+        case configSshOptions of
```
sestrella: @mrkkrp :+1: nice we could easily run commands locally now.

src/System/Hapistrano/Core.hs
```diff
@@ -0,0 +1,98 @@
...
+          Just SshOptions {..} ->
+            ("ssh", [sshHost, "-p", show sshPort, cmd])
+      cmd = renderCommand typedCmd
+      hostLabel =
+        case configSshOptions of
+          Nothing              -> "localhost"
+          Just SshOptions {..} -> sshHost ++ ":" ++ show sshPort
+  liftIO $ do
+    printLine hostLabel
+    putStrLn ("$ " ++ cmd)
```
sestrella: @mrkkrp would be nice to start logging the timestamp as well. We could do that in a different PR.

sestrella: Looking forward maybe we could use a logging library and configure the log level using the command line. But again that's is just a nice to have. No need to do it on this PR. Actually I'm going to create a GH issue.

src/System/Hapistrano/Commands.hs
```diff
@@ -0,0 +1,273 @@
...
+  renderCommand Whoami = "whoami"
+  parseResult Proxy = trim
+
+-- | Specify directory in which to perform another command.
+
+data Cd cmd = Cd (Path Abs Dir) cmd
+
+instance Command cmd => Command (Cd cmd) where
+  type Result (Cd cmd) = Result cmd
+  renderCommand (Cd path cmd) = "cd " ++ quote (fromAbsDir path) ++
```
mrkkrp: This is a good idea, gonig to add this.

PR #53, opened 2017-02-08
Add verification before rollbacks

### LINKED ISSUE #15: Validation before rollback to stable release
Digging a little inside Capistrano source code, I didn't find any kind of validation mechanism that guarantees that you don't rollback to an incomplete revision. The rollback task seems that only changes the current release symlink to an existing revision and perform some rollback operations. Maybe we could use GitHub tags to perform the rollback, so once the deployment is done `hapistrano` should create a new tag with the timestamp we could use an approach like [this](http://casperfabricius.com/site/2009/09/20/manage-and-rollback-heroku-deployments-capistrano-style/).

Some other references:
- [Capistrano flow](http://capistranorb.com/documentation/getting-started/flow/)
- [Capistrano rake tasks](https://github.com/capistrano/capistrano/blob/master/lib/capistrano/tasks/deploy.rake)

mrkkrp:
Close #15.

### DIFF
--- a/spec/System/HapistranoSpec.hs
+++ b/spec/System/HapistranoSpec.hs
@@ -40,6 +40,13 @@ spec = do
         (liftIO . readFile . fromAbsFile) (rpath </> $(mkRelFile "foo.txt"))
           `shouldReturn` "Foo!\n"
 
+    describe "registerReleaseAsComplete" $
+      it "creates the token all right" $ \(deployPath, repoPath) -> runHap $ do
+        let task = mkTask deployPath repoPath
+        release <- Hap.pushRelease task
+        Hap.registerReleaseAsComplete deployPath release
+        (Hap.ctokenPath deployPath release >>= doesFileExist) `shouldReturn` True
+
     describe "activateRelease" $
       it "creates the ‘current’ symlink correctly" $ \(deployPath, repoPath) -> runHap $ do
         let task = mkTask deployPath repoPath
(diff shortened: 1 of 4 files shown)


spec/System/HapistranoSpec.hs
```diff
@@ -40,6 +40,13 @@ spec = do
         (liftIO . readFile . fromAbsFile) (rpath </> $(mkRelFile "foo.txt"))
           `shouldReturn` "Foo!\n"
 
+    describe "registerReleaseAsComplete" $
+      it "it creates the token all right" $ \(deployPath, repoPath) -> runHap $ do
```
sestrella: @mrkkrp duplicated `it`

src/System/Hapistrano.hs
```diff
@@ -75,8 +88,13 @@ rollback
...
 rollback deployPath n = do
-  xs <- genericDrop n <$> deployedReleases deployPath
-  case xs of
+  crs <- completedReleases deployPath
+  drs <- deployedReleases  deployPath
+  -- NOTE If we don't have any completed releases, then perhaps the
+  -- application was used with older versions of Hapistrano that did not
+  -- have this functionality. We then fall back and use collection of “just”
+  -- deployed releases.
+  case genericDrop n (if null crs then drs else crs) of
```
sestrella: @mrkkrp nice we should make this change backward compatible for sure.

mrkkrp:
@sestrella Token file is not generated in release folders. Can you explain why we should modify the log after a rollback and why having a file is better than having a bunch of empty files in a special directory? With tool we have, it's easier to `touch` and check files than to edit a log file.

mrkkrp:
After all, a rollback doesn't change which deploys are successful/complete and which are not.

sestrella:
@mrkkrp got it. I got confused with `deployPath` variable name I thought that one is the path of the current release. So as long as we store the completed releases in a single place I think is fine. LGTM

PR #57, opened 2017-02-18
Use ‘hap.yaml’ to store configuration

### LINKED ISSUE #56: Use YAML to store configuration instead of environment variables
I think there is no particular reason to use environment variables for configuration. YAML is easy to parse with the `yaml` library and I see a number of reasons to switch to it:

1. It will be handier to store the entire configuration in `hap.yaml` file and the application could load it from that file automatically so `source foo &&` part becomes unnecessary and it's easier to call `hap`.

2. We can represent collections/lists of items properly. This is actually necessary for #54, because we would like to allow to specify arbitrary files to copy, and it should be possible to copy several files/directories. How to put this into environment variables? We could use `:` or something as separator like it's done in `PATH`, but this:

   ```yaml
   deploy_path: '/var/projects/foo'
   host: myserver.com
   port: 2222
   revision: origin/master
   restart_command: echo "foo!"
   build_script:
     - stack setup
     - stack build
   copy_verbatim:
     - file-a.txt
     - file-b.txt
   ```

   is much better.

3. As shown in the example above, this allows to put the build script directly into the configuration file instead of keeping it outside. And when we specify path to build script, it's also error-prone in a sense, because it's usually a relative path and resolved path will depend on current working directory (when playing with `hap` locally I was bitten by this a couple times). YAML also handles `#` comments automatically (although it's not a problem anymore, but anyway).

mrkkrp:
Close #56.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Parse ssh port from `PORT` environment variable.
 * Drop support for GHCs older than 7.10 (because Chris Done's `path` does
   not compile with them, see: https://github.com/chrisdone/path/issues/46).
+* Now Hapistrano uses `hap.yaml` file for all its configuration.
 
 ## 0.2.1.2
 
--- a/README.md
+++ b/README.md
@@ -22,57 +22,64 @@ directory marked with a timestamp on the remote host. It creates this
 new directory quickly by placing a git repository for caching purposes
 on the remote server.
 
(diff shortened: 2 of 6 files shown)


sestrella:
@mrkkrp seems that https://github.com/stackbuilders/hapistrano/pull/58 was based on this branch. So I think same comments apply to this PR.

PR #58, opened 2017-02-18
Allow copy precompiled files

### LINKED ISSUE #54: Add the ability to copy compiled files from CI
It's a known fact that with some projects we're running into issues by compiling on server. For example, V********, if compiled with optimizations, exhausts all available memory and can't finish anymore. It may be impossible to deploy it with Hapistrano even without optimizations after a while. I think there are other projects that suffer from this. For these cases, we need an option to copy pre-compiled files from CI or from a local directory in general, because files compiled on CI are often suitable for runnnig on target server. In any case, the feature is nice to have.

mrkkrp:
Close #54. In this version we just allow to copy arbitrary files and directories specified in the configuration. Copying should be quite fast and it's more straightforward than linking across releases, which is probably language-dependent solution (i.e. we know that we want to link `.stack-work`, but with other languages maybe we would want to copy just binary file). In any case the solution implemented here seems to be more general to me.

### DIFF
--- a/app/Config.hs
+++ b/app/Config.hs
@@ -2,7 +2,8 @@
 {-# LANGUAGE RecordWildCards   #-}
 
 module Config
-  ( Config (..) )
+  ( Config (..)
+  , CopyThing (..) )
 where
 
 import Data.Aeson
@@ -28,7 +29,17 @@ data Config = Config
     -- (usually after a deploy or rollback).
   , configBuildScript :: !(Maybe [GenericCommand])
     -- ^ Build script to execute to build the project
+  , configCopyFiles :: ![CopyThing]
+    -- ^ Collection of files to copy over to target machine before building
+  , configCopyDirs :: ![CopyThing]
+    -- ^ Collection of directories to copy over to target machine before building
   } deriving (Eq, Ord, Show)
+
+-- | Information about source and destination locations of a file\/directory
+-- to copy.
+
+data CopyThing = CopyThing FilePath FilePath
+  deriving (Eq, Ord, Show)
 
 instance FromJSON Config where
   parseJSON = withObject "Hapistrano configuration" $ \o -> do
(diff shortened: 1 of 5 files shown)


app/Main.hs
```diff
@@ -58,6 +58,12 @@ optionParser = Opts
   ( long "version"
   <> short 'v'
   <> help "Show version of the program" )
+  <*> strOption
+  ( long "config"
+  <> short 'c'
+  <> value "hap.yaml"
+  <> metavar "PATH"
```
sestrella: @mrkkrp could you add also [showDefault](https://github.com/pcapriotti/optparse-applicative#quick-start) please. In order to print the default value as part of help text.

app/Config.hs
```diff
@@ -0,0 +1,68 @@
...
+-- to copy.
+
+data CopyThing = CopyThing FilePath FilePath
+  deriving (Eq, Ord, Show)
+
+instance FromJSON Config where
+  parseJSON = withObject "Hapistrano configuration" $ \o -> do
+    configDeployPath <- o .: "deploy_path"
+    configHost       <- o .:? "host"
+    configPort       <- o .:? "port"
```
sestrella: @mrkkrp what do you think about setting the default port value at this level:

```Haskell
o .:? "port" .!= 22
```

That way we don't have to deal with the `Maybe` elsewhere.

sestrella: @mrkkrp same for `host` maybe? With localhost as default value.

app/Config.hs
```diff
@@ -0,0 +1,68 @@
...
+    configRepo       <- o .: "repo"
+    configRevision   <- o .: "revision"
+    configRestartCommand <- (o .:? "restart_command") >>=
+      maybe (return Nothing) (fmap Just . mkCmd)
+    configBuildScript <- o .:? "build_script" >>=
+      maybe (return Nothing) (fmap Just . mapM mkCmd)
+    configCopyFiles  <- fromMaybe [] <$> (o .:? "copy_files")
```
sestrella: @mrkkrp maybe we could use Aeson built-in [.!=](http://hackage.haskell.org/package/aeson-1.1.0.0/docs/src/Data-Aeson-Types-FromJSON.html#.%21%3D) instead to set the default value:

```Haskell
o .:? "copy_files" .!= []
```

app/Config.hs
mrkkrp: `localhost` is encoded as `Nothing` argument to `runHapistrano`, so it's more natural to keep this as it is right now.

app/Config.hs
mrkkrp: This is a good idea, I forgot about this helper myself.

sestrella:
@mrkkrp LGTM

PR #60, opened 2017-02-28
Implement concurrent multihost deployment

### LINKED ISSUE #8: Deploy to multiple hosts concurrently
We should be able to deploy to multiple hosts concurrently, eg., a web cluster that all needs to run the same release of the application.

mrkkrp:
Close #8.

### DIFF
--- a/app/Config.hs
+++ b/app/Config.hs
@@ -7,6 +7,8 @@ module Config
 where
 
 import Data.Aeson
+import Data.Function (on)
+import Data.List (nubBy)
 import Data.Maybe (maybeToList)
 import Data.Yaml
 import Path
@@ -43,13 +45,15 @@ data CopyThing = CopyThing FilePath FilePath
 instance FromJSON Config where
   parseJSON = withObject "Hapistrano configuration" $ \o -> do
     configDeployPath <- o .: "deploy_path"
+    let grabPort m = m .:? "port" .!= 22
     host             <- o .:? "host"
-    port             <- o .:? "port" .!= 22
+    port             <- grabPort o
     hs               <- (o .:? "targets" .!= []) >>= mapM (\m -> do
       host' <- m .: "host"
-      port' <- m .:? "port" .!= 22
+      port' <- grabPort m
       return (host', port'))
-    let configHosts = maybeToList ((,) <$> host <*> pure port) ++ hs
+    let configHosts = nubBy ((==) `on` fst)
+          (maybeToList ((,) <$> host <*> pure port) ++ hs)
     configRepo       <- o .: "repo"
     configRevision   <- o .: "revision"
     configRestartCommand <- (o .:? "restart_command") >>=
--- a/app/Config.hs
+++ b/app/Config.hs
@@ -7,6 +7,7 @@ module Config
 where
 
 import Data.Aeson
+import Data.Maybe (maybeToList)
 import Data.Yaml
 import Path
 import System.Hapistrano.Commands
@@ -16,10 +17,8 @@ import System.Hapistrano.Commands
 data Config = Config
   { configDeployPath :: !(Path Abs Dir)
     -- ^ Top-level deploy directory on target machine
-  , configHost :: !(Maybe String)
-    -- ^ Host to deploy to. If missing, localhost will be assumed.
-  , configPort :: !Word
-    -- ^ SSH port number to use, may be omitted
+  , configHosts :: ![(String, Word)]
+    -- ^ Hosts\/ports to deploy to. If empty, localhost will be assumed.
   , configRepo :: !String
     -- ^ Location of repository that contains the source code to deploy
   , configRevision :: !String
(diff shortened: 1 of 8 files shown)


spec/System/HapistranoSpec.hs
```diff
@@ -146,7 +147,18 @@ justExec path cmd' =
...
+runHap m = do
+  let printFnc dest str =
+        case dest of
+          StdoutDest -> putStr str
+          StderrDest -> hPutStr stderr str
+  r <- Hap.runHapistrano Nothing printFnc m
+  case r of
+    Left n -> do
+      expectationFailure ("Failed with status code: " ++ show n)
+      return undefined
```
sestrella: @mrkkrp is this OK? Shouldn't we exit the process with a specific status code?

app/Config.hs
```diff
@@ -44,8 +43,13 @@ data CopyThing = CopyThing FilePath FilePath
 instance FromJSON Config where
   parseJSON = withObject "Hapistrano configuration" $ \o -> do
     configDeployPath <- o .: "deploy_path"
-    configHost       <- o .:? "host"
-    configPort       <- o .:? "port" .!= 22
+    host             <- o .:? "host"
+    port             <- o .:? "port" .!= 22
+    hs               <- (o .:? "targets" .!= []) >>= mapM (\m -> do
+      host' <- m .: "host"
+      port' <- m .:? "port" .!= 22
```
sestrella: @mrkkrp in order to keep consistency between lines 46-47 and lines 49-50. Maybe we could extract out some of the code to separate functions an reuse it consistently on both cases? That way if we decided to update the default values or something like that we do the change on a single place.

app/Config.hs
```diff
@@ -44,8 +43,13 @@ data CopyThing = CopyThing FilePath FilePath
...
+      return (host', port'))
+    let configHosts = maybeToList ((,) <$> host <*> pure port) ++ hs
```
sestrella: @mrkkrp are we merging hosts for backward compatibility only? What do you think about just keeping `targets` to simplify things and prevent misunderstandings from the user perspective?

app/Config.hs
```diff
@@ -16,10 +17,8 @@ import System.Hapistrano.Commands
 data Config = Config
   { configDeployPath :: !(Path Abs Dir)
     -- ^ Top-level deploy directory on target machine
-  , configHost :: !(Maybe String)
-    -- ^ Host to deploy to. If missing, localhost will be assumed.
-  , configPort :: !Word
-    -- ^ SSH port number to use, may be omitted
+  , configHosts :: ![(String, Word)]
```
sestrella: @mrkkrp what about the scenario when a user add the configuration to the same server twice? Then hapistrano is going to trigger two deployments to the same instance concurrently which make cause some race condition issues. I wonder if we could use a `Map String Word` to prevent this to happen, or maybe validate the hosts before starting the deployment process?

app/Main.hs
```diff
@@ -99,44 +102,80 @@ pReleaseFormat = eitherReader $ \s ->
...
+                    (Hap.exec . Hap.MkDir . parent) dpath
+                    Hap.scpDir srcPath dpath
+                  forM_ configBuildScript (Hap.playScript configDeployPath release)
+                  Hap.registerReleaseAsComplete configDeployPath release
+                  Hap.activateRelease configDeployPath release
+                  Hap.dropOldReleases configDeployPath n
+                  forM_ configRestartCommand Hap.exec
+                Rollback n -> do
+                  Hap.rollback configDeployPath n
+                  forM_ configRestartCommand Hap.exec
```
sestrella: @mrkkrp I think is worth it to have at least one end to end test for this new feature.

spec/System/HapistranoSpec.hs
mrkkrp: This is necessary because `expectationFailure` has incorrect type. It should return `IO a`, but it returns `IO ()` instead. This fixed unit type prevents from using it in the same `case` expression with with more general type. The reason `return undefined` is OK and `IO ()` is not OK is that `expectationFailure` always throws so control flow will never get to `return undefined`. It could be just `undefined` or anything at all with general enough type.

app/Config.hs
mrkkrp: `Map String Word` is OK, could be `Set (String, Word)` but then the same host with different ports will be allowed.

app/Config.hs
mrkkrp: Why not...

app/Config.hs
mrkkrp: I wanted to make it backward compatible and I think it's more natural for users who don't care about multi-host deploys to just specify `host` and `port`.

app/Config.hs
mrkkrp: I think `Map` is overkill, `nubBy` will be OK.

sestrella: @mrkkrp yeah I know `Map` is an external dependency from `containers`. So if we could just use what it is on `base` would be better for now.

app/Config.hs
sestrella: @mrkkrp got it but what about just `port`?

spec/System/HapistranoSpec.hs
sestrella: got it

app/Config.hs
mrkkrp: Well, just looks like not a big deal. I'll add something for `port` though.

sestrella:
@mrkkrp LGTM

PR #64, opened 2017-03-30
Fix version display

psibi:
The old way of seeing the version of the program was by doing
something like this:

hap deploy -v

This patch makes it viewable by doing like this (which I feel is a
proper way to do):

hap -v


mrkkrp:
Looks good, thank you!

PR #66, opened 2017-04-01
Add ability to make version control related activities optional

### LINKED ISSUE #65: Is cloning necessary when `copy_files` is included
In our work, we do *most* of the deployment just by copying the binaries.

hapistrano seems to do a bare clone always. Is there any reason why it is done like this ?

psibi:
Fixes #65 

I have documented on the README.md on how to use this parameter. Let me know if you need any clarifications.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -122,17 +122,19 @@ main = do
       exitFailure
     Right C.Config {..} -> do
       chan <- newTChanIO
+      let task rf = Task { taskDeployPath    = configDeployPath
+                         , taskRepository    = configRepo
+                         , taskRevision      = configRevision
+                         , taskReleaseFormat = rf }
       let printFnc dest str = atomically $
             writeTChan chan (PrintMsg dest str)
           hap sshOpts =  do
             r <- Hap.runHapistrano sshOpts printFnc $
               case optsCommand of
                 Deploy releaseFormat n -> do
-                  release <- Hap.pushRelease Task
-                    { taskDeployPath    = configDeployPath
-                    , taskRepository    = configRepo
-                    , taskRevision      = configRevision
-                    , taskReleaseFormat = releaseFormat }
+                  release <- case configVcAction of
+                               True -> Hap.pushRelease (task releaseFormat)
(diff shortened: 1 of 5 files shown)


app/Main.hs
```diff
@@ -122,17 +122,19 @@ main = do
...
           hap sshOpts =  do
             r <- Hap.runHapistrano sshOpts printFnc $
               case optsCommand of
                 Deploy releaseFormat n -> do
-                  release <- Hap.pushRelease Task
-                    { taskDeployPath    = configDeployPath
-                    , taskRepository    = configRepo
-                    , taskRevision      = configRevision
-                    , taskReleaseFormat = releaseFormat }
+                  release <- case configVcAction of
```
juanpaucar: This function is already too long and imperative. Could we change it to a function like `bool` I'm not sure about pattern matching here. What do you think?

\c: @mrkkrp

juanpaucar:
@psibi Great job. Thanks a lot for your contribution.From my side, I just have that minor comment.

app/Main.hs
psibi: @juanpaucar Thanks for the feedback.

Which function is long - Do you mean the `main` function itself ?

One another way of doing this is making `pushRelease` accept the parameter `bool` like you said. But IMO that is messy as we need to check for the bool and do some actions and ignore some actions based on that. Also since `pushRelease` function is exported, it is a backward incompatible change (if that matters.). I felt it is easy to have separate functions to make it easier for readability and maintenance in the long term.

mrkkrp: Looks OK to me. I don't mind "long" functions unless they unnecessarily duplicate something or are hard to read.

mrkkrp:
@psibi Are you using this patched version right now in production? Is it OK for you?

psibi:
@mrkkrp Yeah, I'm using my fork for the last three to four days in our production systems and it has been working good for us.

mrkkrp:
Good, going to merge and publish a new version now.

psibi:
Thanks everyone!

PR #67, opened 2017-04-13
Add mode hard to reset

flandrade:
Hapistrano sets the release to the correct revision by resetting the head of the git repository while the current branch is `master`. However, if you want to deploy using a different branch, which has diverged from master, `git reset origin/<new-branch>` will have unstaged changes after reset.

Let me know if this is the intended behavior. Otherwise, I added `--hard` to the `GitReset` command to avoid having unstaged changes.

### DIFF
--- a/src/System/Hapistrano/Commands.hs
+++ b/src/System/Hapistrano/Commands.hs
@@ -238,6 +238,7 @@ instance Command GitReset where
   type Result GitReset = ()
   renderCommand (GitReset revision) = formatCmd "git"
     [ Just "reset"
+    , Just "--hard"
     , Just revision ]
   parseResult Proxy _ = ()
 


jsl:
Thanks @flandrade! I can see how this would make it work reliably to deploy new branches. However, the `--hard` option is quite a bit more dangerous than a regular `git reset`. From a quick search, I didn't see a way to make Capistrano do this for you - I think you'd have to write a custom command? 

I'd be ok with adding this in our case (I can see how it could be a common and helpful scenario), if it is somehow exposed to the user via the API, and if they know that a potentially destructive thing is going to happen.@mrkkrp do you know if the current design exposes this as optional to the user somehow? If we're changing the `GitReset` to use `--hard`, we should probably also change the name to `GitResetHard` so that the user knows what they're getting when they call this function.

flandrade:
@jsl Yes, I understand it's risky. I think we're using this approach [before](https://github.com/stackbuilders/hapistrano/blob/e68b2ab052d6cf0b9577c9c6bdd45a3affadc039/src/System/Hapistrano.hs#L400), but it's now clear that we need to come out with a safer approach. I'll discuss with @mrkrp. 

Thanks for your input!

juanpaucar:
@flandrade can we make this flag optional and disabled by default?
\cc @jsl @mrkkrp

flandrade:
@jsl @juanpaucar I have discussed with @mrkkrp and we think we can follow Capistrano's approach with `git checkout`. Therefore, as adding `--hard` is not longer necessary, I close this PR. 

I'll work on a fix this week. Thanks.

Issue #73, opened 2017-07-20
Add the latest version of the package to stackage


sestrella:
The latest version of the project uploaded to stackage is 0.2.1.2 but the current project version is 0.3.2.2. Check for possible conflicts with other packages in the latest LTS.

https://www.stackage.org/lts/package/hapistrano

jpvillaisaza:
Here's one of the versions that needs to be modified: fpco/stackage#2559

PR #77, opened 2017-08-08
Loose constraint for `path-io`

juanpaucar:
This fixes: https://github.com/stackbuilders/hapistrano/issues/73

### DIFF
--- a/src/System/Hapistrano.hs
+++ b/src/System/Hapistrano.hs
@@ -8,7 +8,7 @@
 -- Portability :  portable
 --
 -- A module for creating reliable deploy processes for Haskell applications.
-
+{-# LANGUAGE CPP                 #-}
 {-# LANGUAGE OverloadedStrings   #-}
 {-# LANGUAGE RecordWildCards     #-}
 {-# LANGUAGE ScopedTypeVariables #-}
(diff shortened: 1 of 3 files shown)


mrkkrp:
You should not change `stripDir` to `stripProperPrefix` like this. You need to use CPP to choose between the two depending on version of `path`. Otherwise you're making it work only with `path-0.6.0` and later and lose compatibility with older versions.

mrkkrp:
See for example here: https://github.com/mrkkrp/path-io/blob/master/Path/IO.hs#L396-L400.

juanpaucar:
Thanks for the headsup @mrkkrp. I didn't realize [constraints for path](https://github.com/stackbuilders/hapistrano/blob/master/hapistrano.cabal#L49) are oddly written. I'll also change that

mrkkrp:
Uh, yeah, should be `<= 0.7` now I guess, certainly not `<= 6.0`!

src/System/Hapistrano.hs
```diff
@@ -201,7 +200,11 @@ deployedReleases
 deployedReleases deployPath = do
   let rpath = releasesPath deployPath
   xs <- exec (Find 1 rpath :: Find Dir)
+#if MIN_VERSION_path(0,6,0)
```
sestrella: @juanpaucar I think you need to extract this out to one place:

```HS
stripDirs :: FilePath -> [FilePath] -> m [FilePath]
stripDirs path dirs =
#if MIN_VERSION_path(0,6,0)
  mapM (stripProperPrefix path) dirs
#else
  mapM (stripDir path) dirs
#endif
```

src/System/Hapistrano.hs
```diff
@@ -213,7 +216,11 @@ completedReleases
 completedReleases deployPath = do
   let cpath = ctokensPath deployPath
   xs <- exec (Find 1 cpath :: Find File)
+#if MIN_VERSION_path(0,6,0)
```
sestrella: @juanpaucar you should be able to reuse `stripDirs` here

PR #80, opened 2017-08-09
Correct boundaries for old base package

### LINKED ISSUE #79: Compile failures with base-4.7
As can be seen (not anymore, as I corrected the meta-data already, see https://github.com/stackbuilders/hapistrano/issues/79#issuecomment-321195399) on

https://matrix.hackage.haskell.org/package/hapistrano

there's currently compile failures with all 0.3.* releases, i.e.

```
Configuring library for hapistrano-0.3.2.4..
Preprocessing library for hapistrano-0.3.2.4..
Building library for hapistrano-0.3.2.4..

src/System/Hapistrano.hs:38:8:
    Could not find module ‘Numeric.Natural’
    It is a member of the hidden package ‘nats-1.1.1’.
    Perhaps you need to add ‘nats’ to the build-depends in your .cabal file.
    Use -v to see a list of the files searched for.
cabal: Failed to build hapistrano-0.3.2.4 (which is required by exe:hap from hapistrano-0.3.2.4).
```

I'm going to fixup the meta-data (by tightening the lower bound for `base` to `>=4.8`) on Hackage for the affected releases when I get back to a keyboard (unless you beat me to it).

juanpaucar:
This fixes #79 
it is not possible to support versions older than 7.10 since there were  some api changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.3.3.0
+
+* Correct bounds for base. GHC support for versions older than 7.10 was dropped on 0.3.0.0
+* Add `run_locally` to run user defined commands locally before deployment. Thanks to Sibi (GitHub: psibi) for this contribution
+
 ## 0.3.2.4
 
 * Allow time 1.8
--- a/hapistrano.cabal
+++ b/hapistrano.cabal
@@ -1,5 +1,5 @@
 name:                hapistrano
-version:             0.3.2.4
+version:             0.3.3.0
 synopsis:            A deployment library for Haskell applications
 description:
   .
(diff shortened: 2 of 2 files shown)


sestrella:
@juanpaucar what about: 

> To Hapistrano maintainers: you probably need to add a conditional dependency on nats for older base versions that do not have Numeric.Natural in them.

juanpaucar:
@sestrella I tried but, I got this [errors](https://travis-ci.org/stackbuilders/hapistrano/jobs/262819218) because of the api change on errors

juanpaucar:
@sestrella could you take another look, please? What do you think? Should we remove compatibility or make deeper changes on errors catching to support older versions?

juanpaucar:
@sestrella I'm merging this to push the package to hackage. If an issue is opened to support older versions, we might take a look at this again. In any case support for ghc 7.10 was dropped on [version 0.3.0.0](https://github.com/stackbuilders/hapistrano/blob/master/CHANGELOG.md#0300) but boundaries were not properly adjusted

Issue #63, opened 2017-03-29
support ln command on OSX


simonmichael:
This looks pretty good, thanks!

Testing a deployment on OSX failed, /bin/ln doesn't support -T:
```
$ ln -svT /repos/demo.hledger.org/releases/20170329171058/ /repos/demo.hledger.org/current_tmp
ln: illegal option -- T
usage: ln [-Ffhinsv] source_file [target_file]
       ln [-Ffhinsv] source_file ... target_dir
       link source_file target_file
```

jsl:
Thanks @simonmichael for bringing this to our attention. It should work on OS X. We'll take a look!

patrixl:
This affects FreeBSD also (not surprisingly, both are BSD userland), and not just ln, but also mv.

juanpaucar:
@simonmichael @patrixl @jsl Hi everyone, I know this issue has been around for a while but I wanted your opinion on this. Do you think it'd be better to enable another key on the YAML file to tell the *hapistrano* that the target deployment system won't be a linux system and instruct it to use the command versions without unsupported options instead?
 
It would be something like the normal configuration plus another key like `linux` or `bsd` and set accordingly
```yaml
deploy_path: '/var/projects/my-project'
host: myserver.com
port: 2222
repo: 'https://github.com/stackbuilders/hapistrano.git'
revision: origin/master
build_script:
  - stack setup
  - stack build
restart_command: systemd restart my-app-service

linux: false
```

patrixl:
@juanpaucar Hi! Yes a key in the YAML file would probably make everyone's life simpler, I'm fine with that.

juanpaucar:
Sorry for the very late solving of this issue. This has been solved in #86 and it's available starting from hapistrano 0.3.5.0
Thanks @simonmichael and @patrixl for reporting the issue

PR #86, opened 2017-11-27
Add support for deployment on non-GNU/Linux targets

juanpaucar:
Solves #63 

As the issue mentions, some flags are not supported by those UNIX systems which are not Linux. This means that for `mv` and `ln` we can't use `-T` as an argument; and for `readlink` we can't use `-f` in the way we've been doing it.

On this PR, I introduce a new optional key to make explicit the target platform where Hapistrano will deploy. By default, it would asume it's GNU/Linux. On the other hand is explicitly stated, it will now be able to read `linux`, `windows` and any other option will be taken as a BSD system

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,11 +13,13 @@ matrix:
       addons: {apt: {packages: [cabal-install-1.24,ghc-7.10.3],sources: [hvr-ghc]}}
     - env: CABALVER=1.24 GHCVER=8.0.2
       addons: {apt: {packages: [cabal-install-1.24,ghc-8.0.2], sources: [hvr-ghc]}}
-    - env: CABALVER=2.0 GHCVER=8.2.1
-      addons: {apt: {packages: [cabal-install-2.0,ghc-8.2.1], sources: [hvr-ghc]}}
+    - env: CABALVER=2.0 GHCVER=8.2.2
+      addons: {apt: {packages: [cabal-install-2.0,ghc-8.2.2], sources: [hvr-ghc]}}
+    - os: osx
 
 before_install:
-  - export PATH=/opt/ghc/$GHCVER/bin:/opt/cabal/$CABALVER/bin:$PATH
+  - if [[ "$TRAVIS_OS_NAME" != "osx" ]]; then export PATH=/opt/ghc/$GHCVER/bin:/opt/cabal/$CABALVER/bin:$PATH; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update && brew install ghc cabal-install; fi
 
 install:
  - cabal --version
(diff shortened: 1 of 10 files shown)


.travis.yml
```diff
@@ -13,11 +13,15 @@ matrix:
...
-      addons: {apt: {packages: [cabal-install-2.0,ghc-8.2.1], sources: [hvr-ghc]}}
+    - env: CABALVER=2.0 GHCVER=8.2.2
+      addons: {apt: {packages: [cabal-install-2.0,ghc-8.2.2], sources: [hvr-ghc]}}
+    - os: osx
 
 before_install:
-  - export PATH=/opt/ghc/$GHCVER/bin:/opt/cabal/$CABALVER/bin:$PATH
+  - if [[ "$TRAVIS_OS_NAME" != "osx" ]]; then export PATH=/opt/ghc/$GHCVER/bin:/opt/cabal/$CABALVER/bin:$PATH; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install ghc cabal-install; fi
```
CristhianMotoche: Do you think it would be a good idea to concatenate commands in this like?
`if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update && brew install ghc cabal-install; fi`

app/Main.hs
```diff
@@ -121,6 +121,9 @@ main = do
       putStrLn (Yaml.prettyPrintParseException err)
       exitFailure
     Right C.Config {..} -> do
+      unless (configTargetSystem == Windows) $ do
+        putStrLn "Windows is not supported as a target platform for deployment"
```
CristhianMotoche: Do you think we can remove Windows from `TargetSystem` since currently we don't support that platform?

app/Config.hs
```diff
@@ -66,14 +74,22 @@ instance FromJSON Config where
...
+    configTargetSystem <- o .:? "target_system" .!= GNULinux
     return Config {..}
 
 instance FromJSON CopyThing where
   parseJSON = withObject "src and dest of a thing to copy" $ \o ->
     CopyThing <$> (o .: "src") <*> (o .: "dest")
 
+instance FromJSON TargetSystem where
+  parseJSON = withText "target system" $ \ts ->
+    pure $ case ts of
```
CristhianMotoche: What about using a `LambdaCase` here?
```haskell
instance FromJSON TargetSystem where
  parseJSON = withText "target system"
    $ pure . \case
      "linux"   -> GNULinux
      "windows" -> Windows
      _         -> BSD
```

I think the yaml for the configuration will be something like:
```yml
target system: windows
```
Right? Only to be clear.

CristhianMotoche:
@juanpaucar LGTM at now. Good job. I left a few comments. Also, please make sure that the flags when the target system is not `Linux` are working. Perhaps, they depend on an specific version of `mv` or `ln` in `Mac OS`.  Thanks.

CristhianMotoche:
LGTM!

juanpaucar:
@sebashack @CristhianMotoche merging it

PR #92, opened 2018-04-19
Docker image

javcasas:
Implements:
* Dockerfile for building Hapistrano

TODO:
* [x] Figure out a way to tag the image built properly with the version.
  * Done from Docker Hub
* [x] Figure out a way to upload the image to the official Docker registry
  * Done from Docker Hub

Acceptance criteria:
* The image is public and can be used in CI for deploying applications.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,9 @@
 language: c
 
-sudo: false
+sudo: required
+
+services:
+    - docker
 
 cache:
   directories:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: c
 
-sudo: required
+sudo: false
 
 services:
     - docker
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ script:
  - cabal test --show-details=always
  - cabal sdist
  - cabal haddock | grep "100%" | wc -l | grep "4"
- - "docker build . -t stackbuilders/hapistrano:$(cat hapistrano.cabal | grep ^version: | awk '{print $2}')"
+ - if [[ "$TRAVIS_OS_NAME" != "osx" ]]; then docker build .; fi
 
 deploy:
   provider: script
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,14 @@ script:
  - cabal test --show-details=always
  - cabal sdist
  - cabal haddock | grep "100%" | wc -l | grep "4"
+ - "docker build . -t stackbuilders/hapistrano:$(cat hapistrano.cabal | grep ^version: | awk '{print $2}')"
+
+deploy:
+  provider: script
+  script: deploy.sh
+  on:
+    branch: master
+    condition: $GHCVER == 8.2.2
 
 notifications:
   email: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,14 +36,7 @@ script:
  - cabal test --show-details=always
  - cabal sdist
  - cabal haddock | grep "100%" | wc -l | grep "4"
- - if [[ "$TRAVIS_OS_NAME" != "osx" ]]; then docker build .; fi
-
-deploy:
-  provider: script
-  script: deploy.sh
-  on:
-    branch: master
-    condition: $GHCVER == 8.2.2
+ - if [[ "$TRAVIS_OS_NAME" != "osx" ]]; then docker build . -t hapistrano; docker run --rm hapistrano --version; fi
 
 notifications:
   email: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,8 @@ script:
  - cabal test --show-details=always
  - cabal sdist
  - cabal haddock | grep "100%" | wc -l | grep "4"
- - if [[ "$TRAVIS_OS_NAME" != "osx" ]]; then docker build . -t hapistrano; docker run --rm hapistrano --version; fi
+ # There is a single Docker image, there are no variants for different versions of GHC
+ - if [[ "$GHCVER" == "8.0.2" ]]; then docker build . -t hapistrano; docker run --rm hapistrano --version; fi
 
 notifications:
   email: false
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,19 @@
+# Build Hapistrano
+FROM fpco/stack-build as build-env
+
+MAINTAINER Javier Casas <jcasas@stackbuilders.com>
+
+WORKDIR /hapistrano
+COPY . .
+RUN stack install
+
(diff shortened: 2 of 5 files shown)


javcasas:
@seuros The Debian footprint is 22MB. The `hap` binary adds another 50MB or so.

Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+FROM debian:stretch-slim
+RUN apt-get update && \
+    apt-get install -y libgmp-dev && \
+    rm -rf /var/lib/apt/lists/*
+
+COPY --from=build-env /root/.local/bin/hap /bin
+
+# To ensure it works on Debian
+RUN hap --version
+CMD /bin/hap
```
goetzc: I think it's better to use [`ENTRYPOINT`](https://docs.docker.com/engine/reference/builder/#entrypoint). Per the documentation:

> An ENTRYPOINT allows you to configure a container that will run as an executable. 

So we could just execute: `docker run -it --rm hap`

goetzc:
In my system I have:
```
alpine       latest              3fd9065eaf02       3 months ago       4.14MB
debian       stretch-slim        dd99abd0503e       5 weeks ago        55.3MB
debian       9                   da653cee0545       4 months ago       100MB
```

.travis.yml
```diff
@@ -33,6 +36,14 @@ script:
  - cabal test --show-details=always
  - cabal sdist
  - cabal haddock | grep "100%" | wc -l | grep "4"
+ - if [[ "$TRAVIS_OS_NAME" != "osx" ]]; then docker build .; fi
+
+deploy:
```
sestrella: @javcasas I think it would be better to build the images directly in Docker Hub https://docs.docker.com/docker-hub/builds/#create-an-automated-build. We could use GH tags to trigger builds in Docker Hub.

.travis.yml
```diff
@@ -1,6 +1,9 @@
 language: c
 
-sudo: false
+sudo: required
```
juanpaucar: Do you need `sudo`?
I couldn't find a command requiring `sudo` preceding it or is it because of docker?

Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+# Copy Hapistrano to a basic Debian
+FROM debian:stretch-slim
+RUN apt-get update && \
+    apt-get install -y libgmp-dev && \
+    rm -rf /var/lib/apt/lists/*
+
+COPY --from=build-env /root/.local/bin/hap /bin
+
+# To ensure it works on Debian
+RUN hap --version
```
sestrella: @javcasas I don't think you need to run this command during the building process of the image, I'll suggest to run it as part of Travis building process as I mentioned before. What do you think?

.travis.yml
```diff
@@ -33,6 +36,14 @@ script:
  - cabal test --show-details=always
  - cabal sdist
  - cabal haddock | grep "100%" | wc -l | grep "4"
+ - if [[ "$TRAVIS_OS_NAME" != "osx" ]]; then docker build .; fi
```
sestrella: @javcasas would be good to run a command inside the container as part of the building process, so we check the binary is working properly.

```
docker build -t hapistrano
docker run hapistrano --version
```

Dockerfile
```diff
@@ -0,0 +1,28 @@
+# Build Hapistrano
+FROM fpco/stack-build as build-env
+
+MAINTAINER Javier Casas <jcasas@stackbuilders.com>
+
+WORKDIR /hapistrano
+COPY stack.yaml .
+COPY hapistrano.cabal .
+RUN stack build --only-dependencies
```
sestrella: @javcasas in order to download the dependencies you only need `hapistrano.cabal` what about moving `COPY stack.yaml .` below `RUN stack build --only-dependencies` in order to preserve the cache.

Dockerfile
```diff
@@ -0,0 +1,28 @@
+# Build Hapistrano
+FROM fpco/stack-build as build-env
```
sestrella: @javcasas what about using `haskell:8.2` instead?

| Image | Size |
| --- | --- |
| fpco/stack-build:latest | 3 GB |
| haskell:8.2 | 266 MB |

goetzc: Thet Haskell image uses rather old libraries (specially old gcc), and will need a lot of dependencies, but if it works then better so.

Dockerfile
javcasas: Yup, I'll remove it, because it will be checked in CI.

.travis.yml
javcasas: Yup, I think it's better to build on the Docker hub.

.travis.yml
javcasas: Some reference documentation demanded `sudo`, but I'll try again without it.

Dockerfile
javcasas: I'll try that one.

Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+
+# This forces cabal to link Hapistrano and produce an actual binary
+RUN cabal run -- --version
+# Compress the resulting binary
+RUN upx /hapistrano/dist/build/hap/hap
+
+# Copy Hapistrano to a basic Alpine
+FROM alpine:3.7
+
+RUN apk update \
```
seuros: since we have static bin, we don't need these headers

javcasas: Removed files.

juanpaucar:
@javcasas Is there something else preventing you from merging this?

Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+        gmp-dev \
+        gnupg \
+        libffi-dev \
+        linux-headers \
+        upx@testing \
+        zlib-dev
+
+WORKDIR /hapistrano
+
+COPY hapistrano.cabal .
```
goetzc: Better to `git clone` the repo, as we need `.git` when compiling, to fix `hap --version` as currently it doesn't show the commit and branch; there would be also less steps.

javcasas: That would checkout the last version, and would make it difficult to construct older versions. I'll see if I can copy the .git inside in order to be able to fix that.

javcasas:
@juanpaucar I want to test the last bit in order to show the version, but other than that, I'll merge it ASAP. It is a safe-ish change, as it will generate a docker image with the `latest` tag, and only when we tag further versions we will get tagged images.

javcasas:
```
1 jcasas@mac-2017-29:~/projects/hapistrano$ docker run --rm e127e27d106b --version
Hapistrano: 0.3.5.4
branch: docker-image
revision: d59379ac4c4414e56a839caf6243105d86a183e9
```
Awesome!

PR #100, opened 2018-07-24
Bump yaml upper path-io to < 1.5

juanpaucar:


juanpaucar:
@CristhianMotoche Sure, it's also on the commit message
commercialhaskell/stackage#3857

PR #101, opened 2018-08-20
Updated YAML upper bound.

kinorius123:


juanpaucar:
@CristhianMotoche @sestrella @kinorius123 Guys, could you hold for a sec.

I was reviewing if it was worth to create a new version everytime we should into upper bounds issues. So I started using revisions instead?
http://hackage.haskell.org/package/hapistrano-0.3.5.9/revisions/

I only need to make a minor change to the package file using Hackage and then this PR wouldn't need to create a new version but simply loose the constraints.

Do you guys think it's worth to create a new version everytime we have constraints issues? In this case, `yaml` didn't even have a breaking change http://hackage.haskell.org/package/yaml-0.10.0/changelog

sestrella:
@juanpaucar even though, Hackage has its own build matrix, I think is important to push this changes through GH in order to ensure the latest changes are properly tested by Circle CI first before releasing a version to Hackage, additionally, I think is important to keep all package boundaries changes, up to date in the git history.

juanpaucar:
@sestrella No worries, I tested the change locally first. Then, I upgraded the the revision.

I think this pr could have the changelog like

```md
**HEAD
- Bounds for yaml-0.10

**0.3.5.9
- Bounds for path-io
``` 

So that when we finally make a new release, we just move what's in head under a new version.

juanpaucar:
In any case, feel free to merge if you consider so @CristhianMotoche @kinorius123 @sestrella 
I just added a commit to solve something I fixed via revisions and was waiting for another PR

PR #102, opened 2018-09-21
Add support to interpolate ENV variables in a configuration file

sestrella:
### Problem

Sometimes developers need to deploy a feature branch temporarily, therefore providing ENV interpolation support would make it possible to override values in a configuration file. If a configuration file looks as follows:

```yaml
# deploy/staging.yaml
revision: "_env:HAPISTRANO_REVISION:origin/master"
...
```
The revision could be overridden as follows, during the deployment:

```bash
HAPISTRANO_REVISION=origin/feature_branch stack exec hap -- deploy -c deploy/staging.yaml
```

### TODO

- [x] Update README

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 0.3.6.0
+* Add support to interpolate ENV variables in a configuration file.
+
 ## 0.3.5.10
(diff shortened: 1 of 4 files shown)


sestrella:
@juanpaucar done could you please take a second look

PR #103, opened 2018-09-24
Loose constraint for stm and include ghc 8.6

juanpaucar:


.travis.yml
```diff
@@ -20,6 +20,8 @@ matrix:
       addons: {apt: {packages: [cabal-install-2.0,ghc-8.2.2], sources: [hvr-ghc]}}
     - env: CABALVER=2.2 GHCVER=8.4.3
       addons: {apt: {packages: [cabal-install-2.2,ghc-8.4.3], sources: [hvr-ghc]}}
+    - env: CABALVER=2.4 GHCVER=8.6.1
```
sestrella: @juanpaucar not sure about including this as part of this PR since GHC 8.6 is fairly new. Maybe we could add GHC 8.6 to the build matrix under `allow_failures` section, take a look at Yesod as a reference https://github.com/yesodweb/yesod/blob/master/.travis.yml#L90. But for now, I would say let's add this as a separate PR, sounds good to you?

PR #106, opened 2018-10-19
[Closes #105] Read `release-format` and `keep-releases` from the configuration file

### LINKED ISSUE #105: Declarative `hap deploy` configuration
Hapistrano only reads deployment settings like `release-format` and `keep-release` from the CLI.

Taking into account the following configuration file:

```yaml
deploy_path: '/var/projects/my-project'
host: myserver.com
port: 2222
revision: origin/master
build_script:
  - stack setup
  - stack build
restart_command: systemd restart my-app-service
```

If a user required to override the default deploy settings, needs to do that via the CLI as follows:

```sh
hap deploy -k 3 -r long -c hap.yaml
```

The current approach makes it hard to achieve reproducible builds since users have to remember to pass the same settings anytime they invoke `hap deploy`.

Hapistrano should be able to read those settings from the configuration file itself:

```yaml
deploy_path: '/var/projects/my-project'
host: myserver.com
port: 2222
revision: origin/master
release_format: long
keep_releases: 3

build_script:
  - stack setup
  - stack build
restart_command: systemd restart my-app-service
```

That way hapistrano could be invoked from the CLI without passing any extra arguments:

```sh
hap deploy -c hap.yaml
```

sestrella:
Co-authored-by: Esteban Ibarra <eibarra1992@hotmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,11 +72,11 @@ The following parameters are *optional*:
   necessary. The platform where Hapistrano is running won't affect the
   available options for commands (g.e. A Mac deploying to a Ubuntu machine,
   doesn't need this flag)
-* `release_format (optional)` - The release timestamp format, the
+* `release_format` - The release timestamp format, the
   '--release-format' argument passed via the CLI takes precedence over this
   value. If neither CLI or configuration file value is specified, it defaults
   to 'short'
-* `keep_releases (optional)` - The number of releases to keep, the
+* `keep_releases` - The number of releases to keep, the
   '--keep-releases' argument passed via the CLI takes precedence over this
   value. If neither CLI or configuration file value is specified, it defaults
   to '5'
@@ -183,7 +183,7 @@ usually, which is mainly useful for testing.
 
 ## Docker
 
-If you would like to use Docker, there is a lightweight image 
+If you would like to use Docker, there is a lightweight image
 available on [Docker Hub](https://hub.docker.com/r/stackbuilders/hapistrano/).
 
 ## License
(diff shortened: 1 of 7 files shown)


src/System/Hapistrano/Types.hs
```diff
@@ -124,3 +138,13 @@ parseRelease s = (Release ReleaseLong <$> p releaseFormatLong s)
...
+
+-- | Get release format based on the CLI and file configuration values.
+
+fromMaybeReleaseFormat :: Maybe ReleaseFormat -> Maybe ReleaseFormat -> ReleaseFormat
+fromMaybeReleaseFormat cliRF configRF = fromMaybe ReleaseShort (cliRF <|> configRF)
+
+-- | Get keep releases based on the CLI and file configuration values.
+
+fromMaybeKeepReleases :: Maybe Natural -> Maybe Natural -> Natural
+fromMaybeKeepReleases cliKR configKR = fromMaybe 5 (cliKR <|> configKR)
```
juanpaucar: Could you please move this `5` to a constant, please?

spec/System/HapistranoSpec.hs
```diff
@@ -36,6 +39,48 @@ spec = do
         , "cabal install --only-dependencies -j"
         , "cabal build -j" ]
 
+  describe "fromMaybeReleaseFormat" $ do
+    context "when the first value is present" $ do
```
juanpaucar: Could you change the message on the test with the proper names, like `when the command line value is present`, please?.

This applies for all the new tests

README.md
```diff
@@ -72,6 +72,14 @@ The following parameters are *optional*:
   necessary. The platform where Hapistrano is running won't affect the
   available options for commands (g.e. A Mac deploying to a Ubuntu machine,
   doesn't need this flag)
+* `release_format (optional)` - The release timestamp format, the
```
juanpaucar: (optional) isn't required here and below, since this section already states:

> The following parameters are optional

Issue #84, opened 2017-10-10
Trying to deploy to a host using `zsh` as default shell fails


nebtrx:
When you try to deploy to a host using `zsh` as default shell it fails with the following error:
```
zsh:1: no matches found: +refs/heads/*:refs/heads/*
```
That happens because `zsh` doesn't run a command at all when it fails to glob while `bash` tries to run the whole pattern unchanged. There are ways to change the default mode of `zsh`(and other shells having the same behavior) regarding this to match `bash` as explained [here](https://stackoverflow.com/a/15893007/937106) but, it won't work out of the box. 

Do you guys think it would worth to add support for running hapistrano out of the box with `zsh`(and all the shells having the same behavior)? I think it can't be quite useful for those playing with the library locally.

Let me know what you guys think of this. The solution is quite simple, it only requires to properly escape the `*` character so that the globbing would be able to find matches. Allow me to remind you that this isn't a workaround. The globbing fails as well while using `bash` but, its default behavior allows it dodge the ball. 

So, if you agree, I can push a PR adding support for `zsh`(and all the shells having the same behavior) out of the box within minutes

Cheers!

po5i:
![](https://github.trello.services/images/mini-trello-icon.png) [Trying to deploy to a host using `zsh` as default shell fails](https://trello.com/c/cQ5iAUi0/46-trying-to-deploy-to-a-host-using-zsh-as-default-shell-fails)

CristhianMotoche:
My current shell:

```
$ echo $SHELL
/usr/bin/zsh
```

After setting up `ssh` in my personal computer (`Arch OS`):

```
$ systemctl start sshd.service
```
I copied my ssh keys to `~/.ssh/authorized_keys`.

I could replicated with the following `hap.yaml`:
```yaml
deploy_path: '/tmp/hapistrano/'
host: localhost
port: 22
repo: 'https://github.com/stackbuilders/hapistrano.git'
revision: origin/master
build_script:
  - stack setup
  - stack build
linked_files:
  - .env
```


As @nebtrx mentioned, the change is simple:
```
instance Command GitFetch where
  type Result GitFetch = ()
  renderCommand (GitFetch remote) = formatCmd "git"
    [ Just "fetch"
    , Just remote
    , Just "+refs/heads/\\*:refs/heads/\\*" ]
  parseResult Proxy _ = ()
```

Although, we need to configure tests to run on a different shell as well.

nebtrx:
[...]

PR #110, opened 2019-01-04
Support for zsh default shell fails

CristhianMotoche:
Solves #84 
---
@juanpaucar Could you take a look please?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.3.8.1
+### Added
+* Support to deploy to a host that has default `zsh` shell
+
 ## 0.3.8.0
 ### Added
 * `execWithInheritStdout` was added to `System.Hapistrano.Core` to stream output children's
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
-## 0.3.8.1
+## 0.3.9.0
 ### Added
 * Support to deploy to a host that has default `zsh` shell
+* Support to deploy using a different shell. Currently supported: `zsh` and `bash`.
 
 ## 0.3.8.0
 ### Added
--- a/hapistrano.cabal
+++ b/hapistrano.cabal
@@ -1,5 +1,5 @@
 name:                hapistrano
-version:             0.3.8.1
+version:             0.3.9.0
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.3.9.0
```
CristhianMotoche: Not sure which PR will be merged first #109 or this one. So, I write it like that.

hapistrano.cabal
```diff
@@ -1,5 +1,5 @@
 name:                hapistrano
-version:             0.3.8.0
+version:             0.3.9.0
```
juanpaucar: I think we can merge both PRs under the same version

src/System/Hapistrano/Commands.hs
```diff
@@ -245,7 +245,7 @@ instance Command GitFetch where
   renderCommand (GitFetch remote) = formatCmd "git"
     [ Just "fetch"
     , Just remote
-    , Just "+refs/heads/*:refs/heads/*" ]
+    , Just "+refs/heads/\\*:refs/heads/\\*" ]
```
juanpaucar: It's not clear to me this part. Would you mind explain it, please?

CristhianMotoche: I think the explanation about zsh failing due to globbing is clear in the answer that @nebtrx linked in #84: https://stackoverflow.com/questions/15892624/how-to-config-gerrit-refs-in-zsh/15893007#15893007

Basically, `zsh` fails because hapistrano tries to execute:
```
$ git fetch origin +refs/heads/*:refs/heads/*
```
In `zsh` the glob fails in that case unless you escape the `*` properly. In this case, I'm escaping the commands adding a back slash before `*`, I had to add a double back slash `\\` because `\*` is not a valid string in Haskell:

```haskell
repl> putStrLn "\*"

<interactive>:4:12: error:
    lexical error in string/character literal at character '*'
```

hapistrano.cabal
CristhianMotoche: Sounds good to me.

CHANGELOG.md
juanpaucar: You are going to need to rebase the PR

PR #112, opened 2019-02-07
Update Dockerfile

nickovivar:
# Alpine version upgrade
- Upgrade from Alpine 3.7 to Alpine 3.9
- Security updates 
# Add new dependencies to the final image
- Install `ca-certificates` due to an issue in Circle CI at restoring cache. Circle CI demands that root ca-certificates must be available so we can authenticate to the caching servers. 
- Install `git` because it's necessary for the checkout in Circle CI.
screenshot from 2019-02-07 11-39-05

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build Hapistrano
-FROM alpine:3.7 as build-env
+FROM alpine:3.9 as build-env
 
 MAINTAINER Javier Casas <jcasas@stackbuilders.com>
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -42,11 +42,13 @@ RUN cabal build hap
...
 # Copy Hapistrano to a basic Alpine with SSH
-FROM alpine:3.7
+FROM alpine:3.9
 
 RUN apk update \
  && apk add \
-        openssh-client
+        openssh-client \
+        ca-certificates \
+        git
```
CristhianMotoche: @nickovivar Is that why `git` and the other packages wan't installed?

nickovivar: Yes, @CristhianMotoche. To bring a better explanation about this, we are using a multistage Dockerfile. In the first part, we build all the hapistrano project with all dependencies and resources, and, in the second stage, we just get the binary of hapistrano. In this step, we were installing only the OpenSSH-client, but we need some other dependencies that will be useful in Circle CI and others similar platforms.

Issue #120, opened 2019-08-07
Allow to add extra flags to the SSH connection


juanpaucar:
Different servers might have different configurations for the ssh connections, so we might want to add the option for `non-login` and `non-interactive`. We might also consider additional options

juanpaucar:
@speedlight Do you consider that we should add other connection options additional to the previously mentioned options?

speedlight:
> @speedlight Do you consider that we should add other connection options additional to the previously mentioned options?

@juanpaucar I think with this is good to have more control of how ssh session are started.

CristhianMotoche:
Close do to #122

PR #122, opened 2019-10-26
Add SSH args

CristhianMotoche:
To solve #120 

## TODO:
- [x] Refactor a little bit: Change tuple `(,,)` with a data type
- [x] Test the ssh args locally: Testing file example:

```yml
deploy_path: '/tmp/my-project'
host: localhost
port: 22
ssh_args:
 - "-v"
revision: origin/master
build_script:
  - stack setup
  - stack build
restart_command: echo "FINISHED!"`
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ The following parameters are *optional*:
   is useful for testing and playing with `hap` locally).
 * `port` — SSH port number to use. If missing, 22 will be used.
 * `shell` — Shell to use. Currently supported: `zsh` ans `bash`. If missing, `Bash` will be used.
+* `ssh_args` — Optional ssh arguments. By default none are passed.
 * `build_script` — instructions how to build the application in the form of
   shell commands.
 * `restart_command` — if you need to restart a remote web server after a
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ The following parameters are *optional*:
   is useful for testing and playing with `hap` locally).
 * `port` — SSH port number to use. If missing, 22 will be used.
 * `shell` — Shell to use. Currently supported: `zsh` ans `bash`. If missing, `Bash` will be used.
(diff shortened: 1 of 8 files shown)


CristhianMotoche:
Hello @juanpaucar Could you please take a look?

CristhianMotoche:
This is failing due to: #124 I hope we can merge #125 before this.

src/System/Hapistrano/Core.hs
```diff
@@ -110,7 +110,7 @@ getProgAndArgs cmd = do
       Nothing ->
         (renderShell configShellOptions, ["-c", cmd])
       Just SshOptions {..} ->
-        ("ssh", [sshHost, "-p", show sshPort, cmd])
+        ("ssh", sshArgs ++ [sshHost, "-p", show sshPort, cmd])
```
juanpaucar: should `sshArgs` overwrite the `port` option too?

README.md
```diff
@@ -61,6 +61,7 @@ The following parameters are *optional*:
   is useful for testing and playing with `hap` locally).
 * `port` — SSH port number to use. If missing, 22 will be used.
 * `shell` — Shell to use. Currently supported: `zsh` ans `bash`. If missing, `Bash` will be used.
+* `ssh_args` — Optional ssh arguments. By default none are passed.
```
juanpaucar: I would change this text to mention that only the `port` argument is passed to ssh

src/System/Hapistrano/Core.hs
CristhianMotoche: @juanpaucar Good catch! We're lucky :four_leaf_clover: SSH won't complain if it contains a flag duplicated. Also, it seems that it will take the latter rather than the former. Therefore, `sshArgs` won't overwrite `port`. 

I tried: `ssh -p 2222 -p 22 user@127.0.0.1` and it tried to access via `22` rather than `2222`.

Issue #124, opened 2019-10-28
Travis CI failing OSX due to homebrew


CristhianMotoche:
We are using [homebrew](https://github.com/stackbuilders/hapistrano/blob/master/.travis.yml#L29) to install ghc, but it's failing during the [script](https://travis-ci.org/stackbuilders/hapistrano/jobs/603333326#L1881) because it's trying to use Cabal 3.0  and GHC 8.8 (I created a separated [issue for that](https://github.com/stackbuilders/hapistrano/issues/123)).

juanpaucar:
Solved by #126

PR #125, opened 2019-10-28
Fix Circle CI for OSX

### LINKED ISSUE #123: Suppor for GHC 8.8 and Cabal 3.0

CristhianMotoche:
Solves #124 

## Solution explanation

[ghcup](https://gitlab.haskell.org/haskell/ghcup) is recommended to install different versions of GHC and Cabal-install. The usage of this tool is a workaround until we fix #123 (who knows how long that would take :sweat_smile: ).


CristhianMotoche:
Hey @juanpaucar I think this wont be necessary since you're fixing #126 I'll close it.

PR #132, opened 2019-11-06
Add nix support

cptrodolfox:
To build hapistrano using nix:
`nix-build release.nix`
To test hapistrano in an environment
`nix-shell --attr env release.nix`
`stack test`

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,31 @@
+{ mkDerivation, aeson, ansi-terminal, async, base, directory
+, filepath, formatting, gitrev, hspec, hspec-discover, mtl
+, optparse-applicative, path, path-io, process, QuickCheck
+, silently, stdenv, stm, temporary, time, transformers
+, typed-process, yaml, git, zlib, zsh
+}:
+mkDerivation {
+  pname = "hapistrano";
+  version = "0.3.10.0";
+  src = ./.;
+  isLibrary = true;
+  isExecutable = true;
+  enableSeparateDataOutput = true;
+  libraryHaskellDepends = [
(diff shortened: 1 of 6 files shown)


cptrodolfox:
To install in the local store use nix-env -i hapistrano -f release.nix.

default.nix
```diff
@@ -0,0 +1,31 @@
...
+  ];
+  executableHaskellDepends = [
+    aeson async base formatting gitrev optparse-applicative path
+    path-io stm yaml
+  ];
+  testHaskellDepends = [
+    base directory filepath hspec mtl path path-io process QuickCheck
+    silently temporary
+  ];
+  testSystemDepends = [ git zlib zsh ];
```
juanpaucar: I have a question. Since many of the commands are actually a wrapper for `ssh` and `scp`, do we need to add them as dependencies for the package?

cptrodolfox: Those dependencies are for testing, as when nix is building the package it creates a nix-shell (environment) with those packages available. But, when you execute the program with "hap" after installation the binaries available will be the ones on your PATH.

cptrodolfox: So I don't think adding ssh and scp is necessary, unless a test uses those commands.

cptrodolfox:
Created commit with some updates in the stack.yaml to use nix for building as seen in [here](http://docs.haskellstack.org/en/stable/nix_integration/).

cptrodolfox:
It appears that with this change in stack.yaml If you don't have nix installed it will not work.
Please, ignore it until I find a solution.

Issue #116, opened 2019-07-18
Add an option to avoid checkout source code from git 


nickovivar:
We need to add an option to avoid check out the source code during deployment. 
We should choose if we want to check out the code from GH or just move a folder with all the files to deploy the application.

For instance, if a project builds the application in CIrcle CI, hapistrano could move the project folder to the server instead of checking out the code and build the application again in the server.

ibarrae:
Would it make sense to refactor the configuration file to something like:
```yaml
....
external-repository:
  url: 
  revision:

local-directory:
...
```
All these values would be optional, but it would require at least one of them
Then when loading up the configuration, if the `local-directory` is a valid directory it will copy all the files inside of it or it will continue working with `checkout`... and everything as before?

ibarrae:
FYI @juanpaucar, I didn't find any configuration related to "copying a directory" instead of taking the contents from a project with source control. Furthermore, one of the [requirements](https://github.com/capistrano/capistrano/blob/master/README.md#requirements) of using *capistrano* is to have a project that uses source control.

What I _did found_ is that you can have plugins for capistrano, and I found this [one](https://github.com/ekho/capistrano-scm-local) that allows the deployment from a local directory.

Hope this helps :smile:

juanpaucar:
@nickovivar For the scenario that you are describing, you could simply copy the necessary compiled files with 
```yaml
copy_files:
  - src: '/home/stackbuilders/my-file.txt'
    dest: 'my-file.txt'
copy_dirs:
  - src: .stack-work
    dest: .stack-work
```

@ibarrae  I don't think that we should have a breaking change for this or change the current syntax that "deploy yaml" files have. However, I do think that it's a good chance to start supporting other version control sources, including a local one. See #12 

This should be kind of a large change for the library. It should essentially change the commands we execute, depending on the version control system. Once that gets decoupled, the commands for a local copy/paste working as a version control system would be a trivial addition. 

Does it sound reasonable to you? @nickovivar  @ibarrae let me know your thoughts

ibarrae:
I think that using `copy_dirs` won't place the contents of the directory under `<deployment-path>/repo` or at least that's what I understood from @nickovivar request.

> For instance, if a project builds the application in CIrcle CI, hapistrano could move the project folder to the server instead of checking out the code and build the application again in the server.

[...]

PR #135, opened 2020-01-22
Copy directory instead of cloning an entire repository

ibarrae:
A proposal for closing #116.

If this makes sense, could you let me know @juanpaucar, so I can carry on with some tests for backward compatibility and updating the documentation :smile:

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,15 +32,18 @@ filesystem and deletes previous releases to avoid filling up the disk.
 
 ## Usage
 
-Hapistrano 0.3.0.0 looks for a configuration file called `hap.yaml` that
+Hapistrano 0.4.0.0 looks for a configuration file called `hap.yaml` that
 typically looks like this:
 
 ```yaml
 deploy_path: '/var/projects/my-project'
 host: myserver.com
 port: 2222
+# To perform version control operations
 repo: 'https://github.com/stackbuilders/hapistrano.git'
 revision: origin/master
+# To copy the contents of the directory
+local_directory: '/tmp/my-project'
 build_script:
   - stack setup
   - stack build
@@ -50,10 +53,16 @@ restart_command: systemd restart my-app-service
 The following parameters are required:
 
 * `deploy_path` — the root of the deploy target on the remote host.
-* `repo` — the origin repository.
-* `revision` — the SHA1 or branch to deploy. If a branch, you will need to
-  specify it as `origin/branch_name` due to the way that the cache repo is
-  configured.
+* Related to the `source` of the repository, you have the following options:
+  - _Git repository_ — consists of two parameters. When these are set,
+    hapistrano will perform version control related operations.
+    **Note:** Only GitHub is supported.
+    * `repo` — the origin repository.
+    * `revision` — the SHA1 or branch to deploy. If a branch, you will need to
+      specify it as `origin/branch_name` due to the way that the cache repo is
+      configured.
+  * `local_directory` — when this parameter is set, hapistrano will copy the
+    contents of the directory.
 
 The following parameters are *optional*:
 
@@ -204,8 +213,8 @@ available on [Docker Hub](https://hub.docker.com/r/stackbuilders/hapistrano/).
 
 If you want to use Nix for building Hapistrano, the required release.nix and default.nix are available.
 
-For installing the hap binary in your local path: 
-```bash 
+For installing the hap binary in your local path:
+```bash
 nix-env -i hapistrano -f release.nix
 ```
 For developing Hapistrano with Nix, you can create a development environment using:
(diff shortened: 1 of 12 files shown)


src/System/Hapistrano/Types.hs
```diff
@@ -57,15 +60,28 @@ data Config =
     -- ^ How to print messages
     }
 
+-- | The origin of the repository. It can be from a version control provider
+-- like GitHub or a local directory.
+data Repository
```
juanpaucar: A better name for this could be `Source`

src/System/Hapistrano/Types.hs
```diff
@@ -57,15 +60,28 @@ data Config =
...
+  = ExternalRepository
```
juanpaucar: Since we could add support for other sources, this constructor and its fields could be better named after Git ...

app/Config.hs
```diff
@@ -97,8 +97,11 @@ instance FromJSON Config where
     let first Target{..} = host
         configHosts = nubBy ((==) `on` first)
           (maybeToList (Target <$> host <*> pure port <*> pure shell <*> pure sshArgs) ++ hs)
-    configRepo       <- o .: "repo"
-    configRevision   <- o .: "revision"
+        getRepository m =
+              LocalDirectory <$> m .: "local_directory"
+          <|> (m .: "external-repository" >>= \m' -> ExternalRepository <$> m' .: "url" <*> m' .: "revision")
```
juanpaucar: I don't think this should be necessary for now and should be safe to remove it. At least not until adding support for more sources like mercury, etc.

src/System/Hapistrano/Types.hs
ibarrae: I think we could take this in a separate PR, since that involves other changes.

app/Config.hs
ibarrae: Which one you mean it's not necessary? The `external-repository` option?

juanpaucar: I think we can safely remove this whole line until we add support for more sources

src/System/Hapistrano/Types.hs
juanpaucar: Since this PR is introducing these new constructors, then it could be modified now. On second thought, `GitRepository` might be a better name for this constructor and a similar approach could be followed for the fields on the record

ibarrae:
Hey @juanpaucar, I've added some tests and also updated the changelog/readme. I bumped the version to `0.4.0.0` on these files but retained from changing in the `.cabal` in case this is wrong. Let me know your thoughts about this :smile:

README.md
```diff
@@ -50,10 +53,16 @@ restart_command: systemd restart my-app-service
 The following parameters are required:
 
 * `deploy_path` — the root of the deploy target on the remote host.
-* `repo` — the origin repository.
-* `revision` — the SHA1 or branch to deploy. If a branch, you will need to
-  specify it as `origin/branch_name` due to the way that the cache repo is
-  configured.
+* Related to the `source` of the repository, you have the following options:
+  - _Git repository_ — consists of two parameters. When these are set,
```
juanpaucar: Please mark this as default

spec/System/HapistranoConfigSpec.hs
```diff
@@ -0,0 +1,53 @@
...
+    { configDeployPath = $(mkAbsDir "/")
+    , configHosts =
+      [ Target
+          { targetHost = "www.example.com"
+          , targetPort = 22
+          , targetShell = Bash
+          , targetSshArgs = []
+          }
+      ]
+    , configSource = LocalDirectory { localDirectoryPath = $(mkAbsDir "/") }
```
juanpaucar: Why is the default configuration for the test the `local` revision?  I think it should be the other way around. And judging by the parser, that is the case. Am I missing something?

src/System/Hapistrano/Config.hs
```diff
@@ -97,8 +97,10 @@ instance FromJSON Config where
     let first Target{..} = host
         configHosts = nubBy ((==) `on` first)
           (maybeToList (Target <$> host <*> pure port <*> pure shell <*> pure sshArgs) ++ hs)
-    configRepo       <- o .: "repo"
-    configRevision   <- o .: "revision"
+        getRepository m =
```
juanpaucar: this might not just be `getRepository` but source

spec/System/HapistranoConfigSpec.hs
ibarrae: You are totally right, thanks

ibarrae:
Thanks for your feedback @juanpaucar, I bumped the version to `0.4.0.0` and also added a changelog entry. I can't think of anything else to add in this PR. Let me know what you think of that.

PR #137, opened 2020-02-17
Add support for LTS-15.0

### LINKED ISSUE #136: Fails to build with LTS-15.0
LTS [15.0](https://www.stackage.org/lts-15.0) was released yesterday, leading in the following compile error:
```bash
Error: While constructing the build plan, the following exceptions were encountered:

In the dependencies for hapistrano-0.3.10.1:
    path-0.7.0 from stack configuration does not match >=0.5 && <0.7  (latest matching version is 0.6.1)
    path-io-1.6.0 from stack configuration does not match >=1.2 && <1.5  (latest matching version is 1.4.2)
needed since hapistrano is a build target.
```

ibarrae:
Closes #136.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.3.10.2
+### Changed
+* Update upper bounds for `path` and `path-io` packages.
+
 ## 0.3.10.1
 ### Added
 * Update Dockerfile and maintainer.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.3.10.2
+## master
 ### Changed
 * Update upper bounds for `path` and `path-io` packages.
 
--- a/hapistrano.cabal
+++ b/hapistrano.cabal
@@ -1,5 +1,5 @@
 name:                hapistrano
-version:             0.3.10.1
+version:             0.3.10.2
 synopsis:            A deployment library for Haskell applications
 description:
   .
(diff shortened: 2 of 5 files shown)


hapistrano.cabal
```diff
@@ -28,7 +28,7 @@ homepage:            https://github.com/stackbuilders/hapistrano
 bug-reports:         https://github.com/stackbuilders/hapistrano/issues
 build-type:          Simple
 cabal-version:       >=1.18
-tested-with:         GHC==7.10.3, GHC==8.0.2, GHC==8.2.2, GHC==8.4.4, GHC==8.6.5
+tested-with:         GHC==7.10.3, GHC==8.0.2, GHC==8.2.2, GHC==8.4.4, GHC==8.6.5, GHC==8.8.1
```
ibarrae: This was added because we are also testing this GHC version with travis, and it was not part of this file.

spec/System/HapistranoSpec.hs
```diff
@@ -42,7 +43,11 @@ spec = do
            expectedOutput `Hspec.shouldSatisfy` (`isPrefixOf` actualOutput)
   describe "readScript" $
     it "performs all the necessary normalizations correctly" $ do
+#if MIN_VERSION_path_io(1,6,0)
+      let spath = $(mkRelFile "script/clean-build.sh")
+#else
       spath <- makeAbsolute $(mkRelFile "script/clean-build.sh")
+#endif
```
ibarrae: I'm not really confident in this change @juanpaucar but I think it's fine since it's for testing.

The reason behind this is due to the change of types for `AbsPath` in [1.5.0](https://hackage.haskell.org/package/path-io-1.5.0/docs/Path-IO.html#t:AbsPath) from a `kind "*"` to a `Type` in [1.6.0](https://hackage.haskell.org/package/path-io-1.6.0/docs/Path-IO.html#t:AbsPath). If you have a better idea to solve this, I'm more than glad to hear it :smiley:

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.3.10.2
```
juanpaucar: Since #135 is ready too, you could simply set this as `master`. Because it won't be published until #135

```md
## master
### Changed
* My change....

## 0.3.10.1
.
.
.
```

ibarrae: Done! Thanks for the suggestion :smile:

PR #151, opened 2021-03-15
Loose constraint for formatting

juanpaucar:
This solves our part on https://github.com/commercialhaskell/stackage/issues/5686

I'm disabling osx from the testing matrix until we can figure out why it is failing to find those packages on brew. Locally, I can find both packages on brew and they seem to be available: [ghc](https://formulae.brew.sh/formula/ghc) and [cabal](https://formulae.brew.sh/formula/cabal-install). The [error](https://travis-ci.org/github/stackbuilders/hapistrano/jobs/763046417) seems to tell us that it can't find the packages. This will need further investigation. I'll open a separate issue for that.


CristhianMotoche (approved):
Hey @juanpaucar LGTM! If you already test the library in OSX, then I think we're good to go.

Issue #156, opened 2021-05-11
Authenticate pull requests in docker hub


felixminom:
We need to authenticate our pull request in order to test our builds. 

With @sestrella we realized that some of our builds in travis were failing due to a docker hub issue, when looking deeper into it we ended up with this article https://www.docker.com/increase-rate-limits 

In synthesis we need to authenticate our PRs to be able to pull images from docker hub without any limit impediment.

ibarrae:
I can take care of this issue @felixminom @juanpaucar :+1:

sestrella:
@ibarrae I wonder if this is a good time to consider moving to GH actions, so far it looks like there is an official Docker action for this https://github.com/docker/login-action#dockerhub

ibarrae:
> @ibarrae I wonder if this is a good time to consider moving to GH actions, so far it looks like there is an official Docker action for this https://github.com/docker/login-action#dockerhub

That sounds great, tried to keep changes minimal but I'll close the previous PR and work on a new one with GH actions :rocket:

PR #158, opened 2021-05-12
Move to GH actions and add auth step for docker

ibarrae:
This PR moves the CI workflow to GH actions instead of Travis, besides adding the login step for docker in favor of closing #156.

### Caveats
- Travis used to run on ubuntu xenial (16.04). Taking this into account, with GHC 7.10.3, all build were passing correctly, but for the latest OS versions, it **has issues** that were solved in [GHC versions >= 8](https://gitlab.haskell.org/ghc/ghc/-/issues/12759), of course, there's a [workaround](https://github.com/haskell/haskell-platform/issues/304) for it, but for the sake of simplicity, that particular version is run in the old OS version too.


ibarrae:
Hey guys @sestrella @juanpaucar this is ready for review now, the build is failing due to the lack of credentials for login with docker, but besides that all should be OK :+1:

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: Install Zsh
+        run: |
+          if [ "$RUNNER_OS" == "Linux" ]; then
+            sudo apt-get update
+            sudo apt-get install zsh
+          else
+            brew update
+            brew install zsh
+          fi
+      - uses: actions/setup-haskell@v1.1.4
```
sestrella, line 34: @ibarraeit looks like this action is no longer maintained:

image

I think you could use https://github.com/haskell/actions instead

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,64 @@
...
+        id: setup-haskell-cabal
+        with:
+          ghc-version: ${{ matrix.ghc }}
+          cabal-version: "3.0"
```
sestrella, line 38: @ibarrae any particular reason to pin the cabal version? Perhaps we could try the last one available instead, it seems to be the general recommendation.

image

Source:
https://github.com/haskell/actions/tree/main/setup#version-support

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,64 @@
...
+      - run: cabal update --enable-tests
```
sestrella, line 39: @ibarrae I have more experience with `stack` but maybe @jpvillaisaza could provide more input here. AFAIK if you run `cabal configure --enable-tests` first then you don't need to pass this flag to all subsequent commands.

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,64 @@
...
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ${{ steps.setup-haskell-cabal.outputs.cabal-store }}
+            dist-newstyle
+          key: dependencies-${{ runner.os }}-${{ matrix.ghc }}-${{ hashFiles('cabal.project.freeze') }}
+          restore-keys: |
+            dependencies-${{ runner.os }}-${{ matrix.ghc }}-${{ hashFiles('cabal.project.freeze') }}
+            dependencies-${{ runner.os }}-${{ matrix.ghc }}-
+      - run: cabal build --enable-tests
```
sestrella, line 50: @ibarrae just to have better output, I think it would be better to split dependencies installation from the actual compilation of the project, that being said, I wonder if you could install the dependencies first and then compile the project in a separate step.

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,64 @@
...
+        name: Login to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - if: matrix.ghc == '8.0.2'
+        name: Dockerize
+        run: |
+          docker build . -t hapistrano
+          docker run --rm hapistrano --version
```
sestrella, line 64: @ibarrae perhaps I'm missing something, but, it looks like we are building the image but not pushing the result to Docker Hub?

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,64 @@
...
+  build-and-test:
+    strategy:
+      matrix:
+        ghc: ["8.0.2", "8.2.2", "8.4.4", "8.6.5", "8.8.3", "8.10.1"]
+        os: [ubuntu-latest]
+        include:
+          - os: macos-latest
+            ghc: "8.10.1"
+          - os: ubuntu-16.04
+            ghc: "7.10.3"
```
sestrella, line 20: @ibarrae maybe we should consider dropping support for GHC 7.10.3 is quite too old

.github/workflows/build-and-test.yml
ibarrae: Nothing in particular, it just a copy-past of the old travis script, since we're on the plan to update the cabal version, maybe is also a good idea to use `v2-` commands too?

.github/workflows/build-and-test.yml
ibarrae: Totally got me there :disappointed_relieved:, I'm not sure if that would be the case, I've seen some blog posts/CIs on other repos that use the same flags in all commands.

.github/workflows/build-and-test.yml
ibarrae: Yup, that's how the old script was actually. It was never pushing the image so I'm guessing that the purpose was to only check if the image builds properly

.github/workflows/build-and-test.yml
jpvillaisaza: @sestrella If you can `cabal configure` and add some options, that gets stored in a Cabal project file, but you don't need to run `cabal configure`, only if you don't want to specify options later on.

I would leave it with `--enable-tests`, but probably for the following one (`cabal build`), not this one. For this one (`cabal update`), I don't think `--enable-tests` does something, but I'm not sure.

.github/workflows/build-and-test.yml
ibarrae: I'm up for it but I think that would require a bigger change in the code. What I'm trying to say is that removing this from the build matrix is fairly easy but maybe we should also consider CPP statements that we in order to support older versions?

I think that might be worth it's own PR IMHO, just to keep them small :smile:

sestrella: @ibarrae fair enough, would you mind adding a separate ticket to get rid of GHC 7.10.3 at the CI and code level, please?

ibarrae: Sure, I just created a new [issue](https://github.com/stackbuilders/hapistrano/issues/160) for it :+1:

.github/workflows/build-and-test.yml
ibarrae: Great catch! Thanks @sestrella

sestrella (approved):
@ibarrae LGTM

juanpaucar (approved):
LGTM. Thanks for the change, it looks like we now have  warnings being displayed during code review

PR #159, opened 2021-05-13
Update path upper bounds

ibarrae:
Closes [stackage issue](https://github.com/commercialhaskell/stackage/issues/6013)


spec/System/HapistranoSpec.hs
```diff
@@ -43,11 +43,9 @@ spec = do
            expectedOutput `Hspec.shouldSatisfy` (`isPrefixOf` actualOutput)
   describe "readScript" $
     it "performs all the necessary normalizations correctly" $ do
-#if MIN_VERSION_path_io(1,6,0)
-      let spath = $(mkRelFile "script/clean-build.sh")
```
ibarrae, line 47: `spath` is going to be used in `readScript`, which has the following signature:
``` haskell
readScript :: MonadIO m => Path Abs File -> m [GenericCommand]
```

and the signature of `mkRelFile` (after lifting from the Q monad) is
``` haskell
Path Rel File
```

I'm not sure how was this being compiled in the 1st place :thinking: but seems with the last version of `path`, it now triggers an compile exception properly

spec/System/HapistranoSpec.hs
```diff
@@ -43,11 +43,9 @@ spec = do
...
-#else
-      spath <- makeAbsolute $(mkRelFile "script/clean-build.sh")
-#endif
+      spath <- do
+        currentDirectory <- getCurrentDirectory
+        parseAbsFile (currentDirectory ++ "/script/clean-build.sh")
```
sestrella, line 48: @ibarrae using  `getCurrentDirectory` looks good, however, I think it would be better to use [</>](http://hackage.haskell.org/package/path-0.8.0/docs/Path-Posix.html#v:-60--47--62-) to build the path since it is cross-platform. [Path](http://hackage.haskell.org/package/path-0.8.0/docs/src/Path.html) is just an alias that gets resolved to Path.Posix and Path.Windows based on the OS system, so you don't have to worry about using the right path separator if you use `</>`

```
currentDir <- getCurrentDirectory >>= parseAbsFile
return $ currentDir </> "script" </> "clean-build.sh"
```

ibarrae: Nice suggestion, I will apply it asap :rocket:

juanpaucar (approved):
LGTM

Issue #154, opened 2021-03-24
Delete old release only after a successful deployment


sestrella:
Right now hapistrano only deletes old releases after a successful deployment, therefore, old releases could pile up while trying out a deployment script for the first time. Perhaps this behavior should be revisited in order to run the clean-up tasks always  (when the deployment is successful or when it fails)

CristhianMotoche:
Thanks @sestrella for reporting this. Could you please elaborate on this:
> this behavior should be revisited in order to run the clean-up tasks always 
?

I want to make sure I understand correctly. AFAIK, every time you run `hap deploy` it creates a new deploy folder where the build script is executed. Therefore, old releases shouldn't affect that.

Also, I think it's correct to remove old releases only when there is a successful deployment. I don't think we should remove old releases when the release fails. We could run out of old correct releases and that could affect rollbacks.

sestrella:
Sure, let me provide more context. Usually, when a Developer is setting up Hapistrano for the first time in a project, there is a lot of try and error involved, which could easily fill the disk since Hapistrano only deletes the X number of releases specific by `keep_releases` after a successful release. On the other hand, preserving the latest failed release folder is important for troubleshooting purposes, so I think the proposal here is to preserve only the latest failed release and delete the others while keeping the successful releases untouched. So let's consider the following scenario:

```
- S1
- S2 -> current
- F1
```

S = successful release, F = failed release

If another deployment attempt fails, right now the release folder would look like this:

```
- S1
- S2 -> current
- F1
- F2
```
But after this issue, it would look like this:
```
- S1
- S2 -> current
- F2
```

Right now [dropReleases](https://github.com/stackbuilders/hapistrano/blob/ef9df6183dc492a0084ad1b8d3c4a41f24c405ea/app/Main.hs#L168) is only triggered after `activateRelease` if there are no errors.

DavidMazarro:
Hi all, I intend on taking a look at this. I was wondering, is this expected to be the new default behavior of `hap deploy`, or should a new boolean flag be added to the config? (i.e. `delete_failed_releases: yes`)

CristhianMotoche:
Hey @DavidMazarro I prefer to have a boolean flag for this. I think a flag in the command line:
`hap deploy --keep-succeeds` could be better. What do you think?

DavidMazarro:
Hi @CristhianMotoche, I think it's better too, both for compatibility and ability to opt-in/out of this behavior.

PR #174, opened 2022-02-02
Added `keep_one_failed` option that deletes failed releases except for the latest one

DavidMazarro:
Hi everyone. After a couple of weeks (and a lot of back and forth and many questions with @CristhianMotoche, thank you!), I have with an implementation for this feature.

This is provided both as a `hap.yaml` boolean option (`keep_one_failed`) as well as a CLI switch option (`--keep-one-failed` or just `-o`). It works as discussed in #154. One thing to note is that the old failed releases get removed before the `keep_releases` option takes place; I think it makes sense since you wouldn't want to accidentally delete failed releases first with `keep_releases` and then remove some others with `keep_one_failed`, which would also leave you with less releases than the amount you specified with `keep_releases`.

Regarding internal changes, there were a bunch of refactors, some of them in the `Main.hs` module and the `Hapistrano` monad definition, mainly regarding exception handling within `Hapistrano`. The most notable change is the removal of `ctokens`, which were previously used to tag versions as "complete". The new approach proposed in this PR is using a `.hapistrano_deploy_state` within each new release folder, containing `Success` if the deployment succeeded and `Fail` otherwise. The main advantages of this with respect to the `ctokens` approach is that we now don't have to track them separately (as they are included in the releases), and that it's now clear whether a version was deployed before or after creating this feature, in the absence of this file, whereas with `ctokens` releases without one could be either from an old version, or releases that didn't finish the build step properly.

One more thing to note here is that releases are now tagged with respect to whether they could **deploy** properly, as opposed to checking against the **build** (the way it was done with `ctokens`). I don't think this is neither right nor wrong, let me know your thoughts on this.

I haven't changed the app version in the `hapistrano.cabal` file since I'm not sure what's the etiquette on this project, or how exactly the version would change. I'm guessing it would be `0.4.3.1` -> `0.4.4.0`, since it's a minor feature, but please tell me what should I be doing on this regard.

Not sure if I'm leaving anything. Any questions or feedback is more than welcome. 😃

Solves #154.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Compile project
         run: cabal build --enable-tests
       - run: cabal test --enable-tests
-      - run: cabal haddock | grep "100%" | wc -l | grep -G "[45]" # Fixes issue with different haddock coverage with different ghc versions https://github.com/haskell/haddock/issues/123
+      - run: cabal haddock | grep "100%" | wc -l | grep -E "[4-9]|[1-9][0-9]+" # Fixes issue with different haddock coverage with different ghc versions https://github.com/haskell/haddock/issues/123
       - run: cabal sdist
       - if: matrix.ghc == '8.8.4'
         name: Login to Docker Hub
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -15,7 +15,6 @@ import           Data.Version               (showVersion)
 import qualified Data.Yaml.Config           as Yaml
 import           Development.GitRev
 import           Formatting                 (formatToString, string, (%))
-import           Numeric.Natural
 import           Options.Applicative        hiding (str)
 import           Path
 import           Path.IO
@@ -29,21 +28,6 @@ import           System.Hapistrano.Types
 import           System.IO
 
 ----------------------------------------------------------------------------
--- Command line options
-
--- | Command line options.
-
-data Opts = Opts
-  { optsCommand    :: Command
-  , optsConfigFile :: FilePath
-  }
-
--- | Command to execute and command-specific options.
-
-data Command
-  = Deploy (Maybe ReleaseFormat) (Maybe Natural) Bool -- ^ Deploy a new release (with timestamp
-    -- format and how many releases to keep)
-  | Rollback Natural -- ^ Rollback to Nth previous release
 
 parserInfo :: ParserInfo Opts
 parserInfo =
@@ -137,8 +121,8 @@ main = do
                      , taskReleaseFormat = rf }
   let printFnc dest str = atomically $
         writeTChan chan (PrintMsg dest str)
-      hap shell sshOpts =  do
-        r <- Hap.runHapistrano sshOpts shell printFnc $
+      hap shell sshOpts = do
+        r <- Hap.runHapistrano sshOpts shell printFnc Opts{..} C.Config{..} $
           case optsCommand of
             Deploy cliReleaseFormat cliKeepReleases cliKeepOneFailed -> do
               let releaseFormat = fromMaybeReleaseFormat cliReleaseFormat configReleaseFormat
@@ -172,6 +156,7 @@ main = do
               Hap.activateRelease configTargetSystem configDeployPath release
               Hap.dropOldReleases configDeployPath keepReleases
               forM_ configRestartCommand Hap.exec
+              Hap.createHapistranoDeployState configDeployPath release System.Hapistrano.Types.Success
             Rollback n -> do
               Hap.rollback configTargetSystem configDeployPath n
               forM_ configRestartCommand Hap.exec
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -8,10 +8,9 @@ module Main (main) where
 import           Control.Concurrent.Async
 import           Control.Concurrent.STM
 import           Control.Monad
-
-
-
-
+#if !MIN_VERSION_base(4,13,0)
+import           Data.Monoid                ((<>))
+#endif
 import           Data.Version               (showVersion)
 import qualified Data.Yaml.Config           as Yaml
 import           Development.GitRev
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -26,6 +26,8 @@ import qualified System.Hapistrano.Config   as C
 import qualified System.Hapistrano.Core     as Hap
 import           System.Hapistrano.Types
 import           System.IO
+import System.Hapistrano (createHapistranoDeployState)
+import Control.Monad.Error.Class (throwError, catchError)
 
 ----------------------------------------------------------------------------
 
@@ -119,6 +121,10 @@ main = do
   let task rf = Task { taskDeployPath    = configDeployPath
                      , taskSource        = configSource
                      , taskReleaseFormat = rf }
+  let failStateAndThrow e@(_, maybeRelease) =
+        case maybeRelease of
+          (Just release) -> createHapistranoDeployState configDeployPath release Fail >> throwError e
+          Nothing -> throwError e
   let printFnc dest str = atomically $
         writeTChan chan (PrintMsg dest str)
       hap shell sshOpts = do
@@ -160,6 +166,7 @@ main = do
             Rollback n -> do
               Hap.rollback configTargetSystem configDeployPath n
               forM_ configRestartCommand (flip Hap.exec Nothing)
+          `catchError` failStateAndThrow
         atomically (writeTChan chan FinishMsg)
         return r
       printer :: Int -> IO ()
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -113,7 +113,7 @@ data Message
 
 main :: IO ()
 main = do
-  Opts {..} <- execParser parserInfo
+  Opts{..} <- execParser parserInfo
   C.Config{..} <- Yaml.loadYamlSettings [optsConfigFile] [] Yaml.useEnv
   chan <- newTChanIO
   let task rf = Task { taskDeployPath    = configDeployPath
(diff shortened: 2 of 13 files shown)


src/System/Hapistrano.hs
```diff
@@ -225,22 +255,29 @@ deployedReleases
...
+  filterM (
+    fmap ((\bool -> if selectedState == Success then bool else not bool) . stateToBool)
+     . deployState deployPath Nothing
+    ) releases
+  where
+    stateToBool :: Maybe DeployState -> Bool
+    stateToBool mDeployState =
+      case mDeployState of
+        (Just Fail) -> False
+        _ -> True
```
CristhianMotoche, line 280: We can make this shorter this way:
```suggestion
    stateToBool :: Maybe DeployState -> Bool
    stateToBool (Just Fail) = False
    stateToBool _ = True
```

src/System/Hapistrano.hs
```diff
@@ -313,24 +351,25 @@ tempSymlinkPath
...
+  -> Hapistrano (Maybe DeployState) -- ^ Whether the release was deployed successfully or not
+deployState deployPath mWorkingDir release = do
+  parseStatePath <- parseRelFile deployStateFilename
+  actualReleasePath <- releasePath deployPath release mWorkingDir
+  let stateFilePath = actualReleasePath </> parseStatePath
+  doesExist <- exec (CheckExists stateFilePath) (Just release)
+  if doesExist then do
+    deployStateContents <- exec (Cat stateFilePath) (Just release)
+    return $ readMaybe deployStateContents
+  else return Nothing
```
CristhianMotoche, line 372: Hey @DavidMazarro What about adding a different state for this scenario (e.g. `Unknown`):
```hs
data DeployState = Success | Failure | Unknown
```
then:
```suggestion
    fromMaybe Unknown deployStateContents
  else return Unknown
```
WDYT?

src/System/Hapistrano/Commands/Internal.hs
```diff
@@ -190,16 +190,51 @@ instance Command (Find File) where
...
   parseResult Proxy _ = ()
 
+newtype CheckExists =
+  CheckExists
+  (Path Abs File) -- ^ The absolute path to the file you want to check for existence
+
+instance Command CheckExists where
+  type Result CheckExists = Bool
+  renderCommand (CheckExists path) = 
+    "([ -r " <> fromAbsFile path <> " ] && echo True) || echo False"
```
CristhianMotoche, line 208: Hey @DavidMazarro Is `-r` compatible with the shells that we support (`zsh` and `bash`)?

DavidMazarro: I'll admit I didn't initially check, so thanks for the heads-up. Yes, it works on both `zsh` and `bash`.

src/System/Hapistrano.hs
DavidMazarro: I feel a bit embarrased about this one 😅 I'll be changing this, thanks.

flandrade:
@DavidMazarro @CristhianMotoche Could you please review the CI? It seems the build failed

.github/workflows/build-and-test.yml
```diff
@@ -50,7 +50,7 @@ jobs:
       - name: Compile project
         run: cabal build --enable-tests
       - run: cabal test --enable-tests
-      - run: cabal haddock | grep "100%" | wc -l | grep -G "[45]" # Fixes issue with different haddock coverage with different ghc versions https://github.com/haskell/haddock/issues/123
+      - run: cabal haddock | grep "100%" | wc -l | grep -E "[4-9]|[1-9][0-9]+" # Fixes issue with different haddock coverage with different ghc versions https://github.com/haskell/haddock/issues/123
```
jpvillaisaza, line 53: What's this for?

app/Main.hs
```diff
@@ -95,6 +80,11 @@ deployParser = Deploy
             <> help "How many releases to keep, default is '5'"
             )
         )
+  <*> switch
+            ( long "keep-one-failed"
+            <> short 'o'
```
jpvillaisaza, line 85: Is a short option required? I'm just wondering if `-o` (I think normally short for `--output`) is a good choice.

app/Main.hs
```diff
@@ -125,51 +115,62 @@ data Message
...
                      , taskSource        = configSource
                      , taskReleaseFormat = rf }
   let printFnc dest str = atomically $
         writeTChan chan (PrintMsg dest str)
-      hap shell sshOpts =  do
+      hap shell sshOpts = do
         r <- Hap.runHapistrano sshOpts shell printFnc $
           case optsCommand of
-            Deploy cliReleaseFormat cliKeepReleases -> do
+            Deploy cliReleaseFormat cliKeepReleases cliKeepOneFailed ->
```
jpvillaisaza, line 129: I think keeping the previous `-> do` here and removing the `in do` below should make the previous indentation work here, which might make it easier to review.

app/Main.hs
```diff
@@ -125,51 +115,62 @@ data Message
...
-                (Hap.exec . Hap.MkDir . parent) dpath
-                Hap.scpDir srcPath dpath
-              forM_ configLinkedFiles
-                (Hap.linkToShared configTargetSystem rpath configDeployPath)
-              forM_ configLinkedDirs
-                (Hap.linkToShared configTargetSystem rpath configDeployPath)
-              forM_ configBuildScript (Hap.playScript configDeployPath release configWorkingDir)
-              Hap.registerReleaseAsComplete configDeployPath release
-              Hap.activateRelease configTargetSystem configDeployPath release
-              Hap.dropOldReleases configDeployPath keepReleases
```
jpvillaisaza, line 168: Is dropping old releases no longer needed when successful?

src/System/Hapistrano/Config.hs
```diff
@@ -1,3 +1,14 @@
+-- |
+-- Module      :  System.Config
+-- Copyright   :  © 2015-Present Stack Builders
+-- License     :  MIT
+--
+-- Maintainer  :  Juan Paucar <jpaucar@stackbuilders.com>
```
jpvillaisaza, line 6: Was this moved here from somewhere else? If needed, can it be updated (the module name above is incorrect and I'm guessing the maintainer email should be something else)?

src/System/Hapistrano/Types.hs
```diff
@@ -133,12 +136,40 @@ data Release =
...
+data Opts = Opts
+  { optsCommand    :: Command
+  , optsConfigFile :: FilePath
+  }
+
+-- | Command to execute and command-specific options.
+
+data Command
+  = Deploy (Maybe ReleaseFormat) (Maybe Natural) Bool -- ^ Deploy a new release (with timestamp
+    -- format and how many releases to keep)
```
jpvillaisaza, line 169: Looks like this one is missing mentioned the new `Bool`.

jpvillaisaza (commented):
I think this looks good overall. I'm not very familiar with the Hapistrano code, so I added a few questions to better understand the changes.

jpvillaisaza:
I forgot to ask something else. Do you know if there's a test doing what https://github.com/stackbuilders/hapistrano/issues/154#issuecomment-971697101 describes? Or could it be added?

.github/workflows/build-and-test.yml
DavidMazarro: It's explained here: https://github.com/stackbuilders/hapistrano/pull/175#issuecomment-1029203501

src/System/Hapistrano/Config.hs
DavidMazarro: Yeah the module name is wrong, thanks for the heads-up. I didn't want to change the maintainer; if @juanpaucar is no longer the maintainer, who should be there?

app/Main.hs
DavidMazarro: Oops, that was an oversight. Will be fixed in next commit, thanks 👍

app/Main.hs
DavidMazarro: The reason I'm doing this is because I want the `failStateAndThrow` exception handler to be in the outer scope to catch the exceptions that happen within the `do` you mention; otherwise it would remain in the inner scope and can't be used outside.

DavidMazarro:
> I forgot to ask something else. Do you know if there's a test doing what [#154 (comment)](https://github.com/stackbuilders/hapistrano/issues/154#issuecomment-971697101) describes? Or could it be added?

There's not, I'll see if I can add one.

.github/workflows/build-and-test.yml
jpvillaisaza: Sounds good, looks like this shouldn't be here once the branch is updated then.

src/System/Hapistrano/Config.hs
jpvillaisaza: I don't know, maybe @CristhianMotoche knows.

app/Main.hs
DavidMazarro: I'll remove it then. If someone asks for it in the future, we can just add it 🙂

app/Main.hs
jpvillaisaza: Can it be moved to around line 126 then (same level as `hap` and `printFnc`)?

DavidMazarro: I don't think so, because this can only work if we have been asked to run the `deploy` command, that's why it's within the case-splitting.

app/Main.hs
jpvillaisaza: Thanks. I guess it would be good to have a test for the logic here so that it doesn't change. That can be an issue created as outcome of this PR since it's probably out of scope.

DavidMazarro: Agreed, will write up an issue about this after this PR is completed, thanks for the suggestion.

DavidMazarro:
@jpvillaisaza I just added the test that makes sure the program behaves as explained in https://github.com/stackbuilders/hapistrano/issues/154#issuecomment-971697101.

src/System/Hapistrano/Config.hs
CristhianMotoche: Hey @DavidMazarro @jpvillaisaza I'm the maintainer at the moment. You can change the name. Thanks.

CristhianMotoche (approved):
@DavidMazarro LGTM! Thanks for the contribution. I think we can prepare a release for this later. I would like to include another change before the next release.

PR #175, opened 2022-02-03
Changed command at the Haddock step

DavidMazarro:
Let's try this, since this way we can get more information about what's happening during this step and will make debugging failing Haddock buillds easier (if it works).

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Compile project
         run: cabal build --enable-tests
       - run: cabal test --enable-tests
-      - run: cabal haddock | grep "100%" | wc -l | grep -G "[45]" # Fixes issue with different haddock coverage with different ghc versions https://github.com/haskell/haddock/issues/123
+      - run: cabal haddock -v
       - run: cabal sdist
       - if: matrix.ghc == '8.8.4'
         name: Login to Docker Hub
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -50,7 +50,7 @@ jobs:
       - name: Compile project
         run: cabal build --enable-tests
       - run: cabal test --enable-tests
-      - run: cabal haddock | grep "100%" | wc -l | grep -G "[45]" # Fixes issue with different haddock coverage with different ghc versions https://github.com/haskell/haddock/issues/123
+      - run: cabal haddock -v
```
CristhianMotoche, line 53: Hey @DavidMazarro That will help to know the results from `cabal haddock`. However, the purpose of this task was to ensure that the library always provides enough documentation (at least 4 to 5 modules with a 100% of documentation). We should preserve that.

DavidMazarro: Ah, I didn't know that, and this made me realize exactly what was the issue with the CI pipeline in #174: there are 6 modules with 100% documentation, and this is looking for 4 or 5 exactly. We can preserve that check, but we would need to change it somehow, IMO.

DavidMazarro:
I have changed the purpose of the PR; now there's no `-v` option in the Haddock step, but I've changed the regex used to check for test coverage, which only allowed for exactly 4 or 5 modules with 100% coverage, to one that allows for a number of 4 modules or greater with 100% coverage.

CristhianMotoche (approved):
@DavidMazarro Thanks! :shipit:

Issue #176, opened 2022-02-04
Make Haddock step in the CI pipeline clearer in case of failure


DavidMazarro:
As a newcomer developer to the project, after proposing #174 I kept scratching my head as to why the CI checks were failing at the Haddock step with an exit code 1 and no extra details provided. It turns out that the Haddock step:
https://github.com/stackbuilders/hapistrano/blob/db27a71f3d903c2e1b0e78e6d700e04012bfab28/.github/workflows/build-and-test.yml#L53
was checking via `grep` with a regex that the number of modules with 100% documentation was exactly 4 or 5, but in my PR there were 6, so it was failing at the step. This is now being fixed by #175, but I think this kind of confusion could happen to some newcomer in the future.

The two alternatives that come to my mind to make more explicit on what failed would be:
- Rewrite the step in a way that the minimum number of 4 modules with 100% documentation is preserved, maybe without using the `wc -l | grep` trick, or
- Leave the step as-is, but add a message in the Action so that it suggests what might have happened (e.g.: "process failed with exit code 1. Have you checked that the minimum of 4 modules with 100% documentation is fulfilled?")

PR #179, opened 2022-02-09
Added name to Haddock step

DavidMazarro:
The CI step involving Haddock was displaying:
> **Run # Fixes issue with different haddock coverage with different ghc versions https://github.com/haskell/haddock/issues/123**

after #176, since no name was given to the step. This PR is fixing that.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -50,6 +50,7 @@ jobs:
       - name: Compile project
         run: cabal build --enable-tests
       - run: cabal test --enable-tests
+      - name: Generate module documentation with Haddock
       - run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -50,6 +50,7 @@ jobs:
       - name: Compile project
         run: cabal build --enable-tests
       - run: cabal test --enable-tests
+      - name: Generate module documentation with Haddock
       - run: |
```
CristhianMotoche, line 54:
```suggestion
      - name: Generate module documentation with Haddock
        run: |
```
I think `name` and `run` are part of the same object, aren't them?

DavidMazarro: Oops, my bad, you're right. Fixing this right now.

CristhianMotoche (approved):
LGTM!

Issue #181, opened 2022-03-30
Separate `build-and-test` from the release process


CristhianMotoche:
Currently, there is one single workflow file that does everything. Unfortunately, this causes some issues when the contribution comes from an external members because the step where hapistrano is built on a docker container fails. Let's separate [these lines](https://github.com/stackbuilders/hapistrano/blob/db27a71f3d903c2e1b0e78e6d700e04012bfab28/.github/workflows/build-and-test.yml#L55-L65) from the main workflow.

sestrella:
Solved by #183

PR #183, opened 2022-03-31
[#181] Separate build-and-test from the release process 

sestrella:
**Improvements**
- Docker caching layer: Building the Docker image now takes ~2 mins with cache, a full run takes ~ 8 mins (without cache)
- Update Dockerfile base version: Update from Alpine 3.13 and 3.15, as well as make some changes to the Dockerfile in order to cache the dependencies installed
- Parallel execution: Now the regular Haskell build runs in parallel with Docker, before it was a sequential process


.github/workflows/ci.yml
```diff
@@ -0,0 +1,84 @@
...
+            dist-newstyle
+      - run: cabal build --only-dependencies
+      - run: cabal build
+      - run: cabal test
+      - run: |
+          # Fixes issue with different haddock coverage with different ghc versions https://github.com/haskell/haddock/issues/123
+          cabal haddock | grep "100%" | wc -l | grep -E "[4-9]|[1-9][0-9]+" ||
+          (echo "Haddock failed with exit code 1. Have you checked that the minimum of 4 modules with 100% documentation is fulfilled?" && false)
+
+  docker:
```
CristhianMotoche, line 64: @sestrella This looks nice. Where do you specify that this job will be executed only when a release is happening? :thinking:

sestrella: @CristhianMotoche this job mimics the [existing workflow](https://github.com/stackbuilders/hapistrano/blob/master/.github/workflows/build-and-test.yml#L59-L69), in that sense, the current workflow only builds and runs the image, it doesn't interact with Docker Hub therefore, the `docker/login-action` was not required at all

CristhianMotoche (approved):
@sestrella LGTM!

PR #187, opened 2022-04-08
[#188] Push image to a Docker registry

### LINKED ISSUE #188: Automatically push image to Docker registry
Pushing Docker images directly from the CI in response to tag events, would be helpful in order to keep the image up to date

sestrella:
Fixes #188

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -47,7 +47,7 @@ jobs:
           cache-from: type=local,src=/tmp/.buildx-cache
           cache-to: type=local,dest=/tmp/.buildx-cache-new,mode=max
           load: true
-      - run: docker run stackbuilders/hapistrano --version
+      - run: docker run ${{ env.IMAGE_NAME }} --version
       - uses: docker/build-push-action@v2
         with:
           context: .
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -29,6 +29,7 @@ jobs:
       - uses: docker/metadata-action@v3
         with:
           images: stackbuilders/hapistrano
+        id: meta
       - uses: docker/build-push-action@v2
         with:
           context: .
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,49 @@
+---
+name: Docker
+
+on: push
+  # create:
+  #   tags:
+  #     - v*
+  #
+
+concurrency:
+  group: ${{ github.ref }}-docker
+  cancel-in-progress: true
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: docker/setup-buildx-action@v1
+      - uses: actions/cache@v3
+        with:
+          key: v1-${{ runner.os }}-buildx-${{ github.sha }}
+          restore-keys: v1-${{ runner.os }}-buildx-
+          path: /tmp/.buildx-cache
+      - uses: docker/login-action@v1 
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - uses: docker/metadata-action@v3
+        with:
+          images: stackbuilders/hapistrano
+      - uses: docker/build-push-action@v2
+        with:
+          context: .
+          load: true
+          cache-from: type=local,src=/tmp/.buildx-cache
+          cache-to: type=local,dest=/tmp/.buildx-cache-new,mode=max
+      - run: docker run hapistrano --version
+      - uses: docker/build-push-action@v2
+        with:
+          context: .
+          push: true
+          cache-from: type=local,src=/tmp/.buildx-cache
+          cache-to: type=local,dest=/tmp/.buildx-cache-new,mode=max
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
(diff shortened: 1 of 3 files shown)


.github/workflows/docker.yml
```diff
@@ -0,0 +1,67 @@
...
+      packages: write
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: docker/setup-buildx-action@v1
+      - uses: actions/cache@v3
+        with:
+          key: v1-${{ runner.os }}-buildx-${{ github.sha }}
+          restore-keys: v1-${{ runner.os }}-buildx-
+          path: /tmp/.buildx-cache
```
CristhianMotoche, line 31: @sestrella I don't think we'll push a tags to create new releases very frequently. Out of curiosity, how long is the cache available to be used in GH Actions?

.github/workflows/docker.yml
```diff
@@ -0,0 +1,67 @@
...
+          context: .
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          cache-from: type=local,src=/tmp/.buildx-cache
+          cache-to: type=local,dest=/tmp/.buildx-cache-new,mode=max
+          load: true
+      - name: Smoke test
+        run: docker run ${{ env.IMAGE_NAME }} --version
+      - name: Pushing image
+        if: contains(github.ref, 'refs/tags/v')
```
CristhianMotoche, line 53: @sestrella Oh, I see our plan is to build on every commit push and push the image when the ref is a tag. That makes sense. Although, I think we should also avoid the step `docker/login-action@v1` on every push. I guess the builds will fail when users out of the organization do a contribution since they're not part of our GH teams.

CristhianMotoche (commented):
@sestrella LGTM! I left a question to know if external contributors will get blocked in the docker build step. Please, let me know your thoughts.

.github/workflows/docker.yml
sestrella: @CristhianMotoche not really, since pushing tags requires write privileges to the repo, therefore, it can’t only be done by admins and maintainers, I don't think this is going to be an issue for external contributors.

sestrella: Good catch, about `docker/login-action@v1` I'm going to update that

CristhianMotoche: Hey @sestrella I think my previous comment was a bit confusing. I was referring to push commits from external contributors since I see that the `docker/login-action@v1` step is executed on every commit push. Then, it could fail for external contributors, couldn't it?

.github/workflows/docker.yml
sestrella: @CristhianMotoche here are some insights:

**Without cache:**

image

**With cache:**

image

CristhianMotoche: Ignore this comment, too.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,67 @@
+# The present workflow was made based on the following references:
+# - https://evilmartians.com/chronicles/build-images-on-github-actions-with-docker-layer-caching
+# - https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md
+---
+name: Docker
+
+on: push
```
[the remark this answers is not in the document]
sestrella: The idea is to build the image always to catch errors and only pushing on tag creation events.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,81 @@
...
+          labels: ${{ steps.meta.outputs.labels }}
+          cache-from: type=local,src=/tmp/.buildx-cache
+          cache-to: type=local,dest=/tmp/.buildx-cache-new,mode=max
+          push: true
+      - run: |
+          # this is a workaround, check the following issues:
+          # - https://github.com/docker/build-push-action/issues/252
+          # - https://github.com/moby/buildkit/issues/1896
+          rm -rf /tmp/.buildx-cache
+          mv /tmp/.buildx-cache-new /tmp/.buildx-cache
```
CristhianMotoche, line 81: @sestrella I see some duplicated steps between `build` and `publish`. Is it possible to use [node anchors](https://en.wikipedia.org/wiki/YAML#Advanced_components) and references (or [merge](https://yaml.org/type/merge.html)) to make it DRY?

sestrella: @CristhianMotoche unfortunately, YAML anchors are [not supported](https://github.community/t/support-for-yaml-anchors/16128) yet, on the other hand, there is something called [composite run steps](https://github.blog/changelog/2020-08-07-github-actions-composite-run-steps/) but it involves setting up a new repo which adds a level of indirection, so I think some code duplication in the context of this workflow is OK. What do you think?

CristhianMotoche: Too bad anchors are not supported yet. I don't think we should complicate things with the _composite run steps_. Let's keep it simple.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,81 @@
...
+
+  publish:
+    if: contains(github.ref, 'refs/tags/v')
+    runs-on: ubuntu-latest
+    needs: [build]
```
CristhianMotoche, line 46: @sestrella Is it possible to add something like this for the `build` step but referring to `build` job from the `ci` workflow? In that way, if the CI workflow fails then this workflow won't run.

CristhianMotoche (approved):
@sestrella LGTM! I left a last question. Although, I don't think it's a blocker. Thanks for the contribution!

PR #190, opened 2022-04-12
[FIX] Change version to 0.4.4.0

CristhianMotoche:
# Changes

Hapistrano is mainly used as a CLI and its CLI didn't change to request mandatory
field or flags. Therefore, I changed the next version to `0.4.4.0`.


sestrella (approved):
@CristhianMotoche LGTM

PR #192, opened 2022-04-20
Fix build cache issues

sestrella:


.github/workflows/ci.yml
```diff
@@ -25,34 +26,29 @@ jobs:
...
-      - run: cabal build --only-dependencies
+          restore-keys: ${{ runner.os }}-${{ matrix.ghc }}-
       - run: cabal build
       - run: cabal test
-      - run: |
-          # Fixes issue with different haddock coverage with different ghc versions https://github.com/haskell/haddock/issues/123
-          cabal haddock | grep "100%" | wc -l | grep -E "[4-9]|[1-9][0-9]+" ||
-          (echo "Haddock failed with exit code 1. Have you checked that the minimum of 4 modules with 100% documentation is fulfilled?" && false)
+      - run: cabal haddock
+      - run: cabal sdist
```
CristhianMotoche, line 54: @sestrella Why do we need this step?

sestrella: @CristhianMotoche it is not strictly necessary right now, but the output produced by this command could be upload to Hackage

> cabal sdist [FLAGS] [TARGETS] takes the crucial files needed to build TARGETS and puts them into an archive format ready for upload to Hackage. These archives are stable and two archives of the same format built from the same source will hash to the same value.

https://cabal.readthedocs.io/en/latest/cabal-commands.html?highlight=sdist#cabal-sdist

.github/workflows/ci.yml
```diff
@@ -25,34 +26,29 @@ jobs:
...
             dist-newstyle
-      - run: cabal build --only-dependencies
+          restore-keys: ${{ runner.os }}-${{ matrix.ghc }}-
       - run: cabal build
       - run: cabal test
-      - run: |
-          # Fixes issue with different haddock coverage with different ghc versions https://github.com/haskell/haddock/issues/123
-          cabal haddock | grep "100%" | wc -l | grep -E "[4-9]|[1-9][0-9]+" ||
-          (echo "Haddock failed with exit code 1. Have you checked that the minimum of 4 modules with 100% documentation is fulfilled?" && false)
+      - run: cabal haddock
```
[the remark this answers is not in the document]
sestrella: I think I removed that by accident, let me put it back

CristhianMotoche (approved):
:+1: LGTM! :shipit:

PR #196, opened 2022-05-12
Add CI workflow step to publish on Hackage on tag push

### LINKED ISSUE #195: Add CI workflow step to publish on Hackage on tag push
Right now, Hapistrano is on version `v0.4.4.0` as per the [latest release tag](https://github.com/stackbuilders/hapistrano/releases/tag/v0.4.4.0). However, the latest version that is on [Hackage](https://hackage.haskell.org/package/hapistrano) is `v0.4.3.0`. Since we are doing improvements to the workflow (e.g. #194 which aims to add static binary release generation to the CI workflow), it would be fitting to try adding a step that releases the latest version of the library to Hackage so that it's consistent with the GitHub repo. This should be done when a [new tag](https://github.com/stackbuilders/hapistrano/tags) is pushed.

An example for how to do this can be seen in the [hackage.yml](https://github.com/haskell/haskell-language-server/blob/master/.github/workflows/hackage.yml) workflow for the [HLS](https://github.com/haskell/haskell-language-server/) project.

DavidMazarro:
Closes #195.


.github/workflows/ci.yml
```diff
@@ -51,3 +52,38 @@ jobs:
...
+      - name: Create folder for artifacts
+        id: artifact
+        run: |
+          mkdir artifact
+          mkdir artifact/${{ matrix.os }}
+          mkdir artifact/${{ matrix.os }}/${{ matrix.ghc }}
+          echo ::set-output name=directory::'artifact/${{ matrix.os }}/${{ matrix.ghc }}'
+
+      - name: Generate package ready for Hackage distribution
+        run: cabal sdist --output-dir ${{ steps.artifact.outputs.directory }}
```
sestrella, line 65: @DavidMazarro what do you think about using `cabal sdist` default directory here? I think the default directory is `dist-newstyle/sdist`, that way you don't need to create the directory based on `{{ matrix.os }}` and `{{ matrix.ghc }}`

.github/workflows/ci.yml
```diff
@@ -51,3 +52,38 @@ jobs:
...
+
+      - name: Upload the artifact to GitHub
```
sestrella, line 67: @DavidMazarro actually I just realize that [tfausak/strive](https://github.com/tfausak/strive/blob/main/.github/workflows/ci.yaml) calls `cabal update` directly in the context of `release` job without calling `haskell/actions/setup` first, so my guess is that `ubuntu-latest` runner already has `cabal` pre-installed? If that is the case, I think we could get rid of the whole upload-artifact/download-artifact thing and call `cabal sdist` and `cabal upload` directly at the `release` job:

```yml
release:
  ...
  steps:
    - uses: actions/checkout@v3
    - run: cabal sdist
    - run: cabal upload ...
```

DavidMazarro:
Also, another thing to consider: the [Hackage tips](https://hackage.haskell.org/upload) for uploading a package mention that it's advisable to upload the Haddock docs along with the release, instead of relying on their server generating those docs for us, see:
> The hackage-server attempts to build documentation for library packages, but this can fail. Maintainers can generate their own documentation and upload it by using something along the lines of the shell script below [...]
Given that we are generating those docs during our CI workflow, should we try doing that too?

sestrella:
@DavidMazarro good idea! Yes, I think we should try to generate the docs on our end (CI) before uploading a tarball to Hackage

.github/workflows/ci.yml
```diff
@@ -51,3 +52,18 @@ jobs:
...
+    steps:
+      - name: Generate package ready for Hackage distribution
+        run: cabal sdist
+
+      - name: Upload package to Hackage
+        run: |
+          cabal upload --publish \
+            --username '${{ secrets.HACKAGE_USERNAME }}' \
+            --password '${{ secrets.HACKAGE_PASSWORD }}' \
+            dist-newstyle/sdist/hapistrano-${{ github.event.release.tag_name }}.tar.gz
```
sestrella, line 69: @DavidMazarro I think you can get rid of `github.event.release.tag_name` here in favor of `*`, something like this:

`dist-newstyle/sdist/hapistrano-*.tar.gz`

DavidMazarro: We can do this, it's true that it will be simpler. I was a bit concerned about compatibility because glob expressions don't behave consistently across different shells; but I guess if it's always going to be performed under `ubuntu-latest` without any weird tweaks that shouldn't be an issue.

DavidMazarro:
@sestrella I have just included two steps in the `release` job for the docs that will be uploaded to Hackage along with the package the notes at the bottom of [these instructions](https://hackage.haskell.org/upload). Please check it out 🙂

sestrella (approved):
@DavidMazarro nice job! LGTM

PR #199, opened 2022-05-27
Refactoring CI workflows

sestrella:
**Overview**

- Build (triggered on every push)
  - [x] Build and test Haskell code
  - [x] Build Docker image
- Pre-Release (triggered by a tag)
  - [x] Creates a GH draft release with a static binary
- Release (triggered after making the release public)
  - [x] Upload the Docker image to ghcr.io
  - [x] Upload the package and docs to Hackage


.github/workflows/README.md
```diff
@@ -0,0 +1,23 @@
...
+## Events
+
+```mermaid
+sequenceDiagram
+  participant Build
+  participant Draft
+  participant Release
+  Build->>Draft: Tag created
+  Draft->>Release: Publish release
+```
```
DavidMazarro, line 23: The diagram isn't a 100% clear for me. I understand that the arrows going to the right mean moving onto those CI workflows once the event in the arrow happens. However, I'm confused by the fact that the workflows are repeated at the top and at the bottom of the diagram; and to nitpick some more, it feels like all steps are starting if you start reading the diagram from the top, and then during the build step if a tag is created, we move to the draft step mid-execution, which is quite different. I'm thinking something along these lines might be clearer to what we're trying to convey:
```suggestion
```mermaid
  flowchart LR
      Build --> tag{{Is a tag created on GH?}} -- Yes --> Draft
      tag -- No --> endci[End CI]
      Draft --> release{{Is a release published on GH?}} -- Yes --> Release
      release -- No --> endci
      Release --> endci
```

This is how that diagram would look:
Screenshot 2022-06-09 at 11 39 50

Let me know your thoughts on this matter 🙂

.github/workflows/draft.yml
```diff
@@ -0,0 +1,60 @@
...
+            --volume $PWD/bin:/root/bin \
+            ${{ github.repository }} \
+            /usr/local/bin/hap \
+            /root/bin/hap-${{ github.ref_name }}-linux-x86_64-bin
+      - name: Create draft release
+        uses: softprops/action-gh-release@v1
+        with:
+          files: |
+            bin/hap-*
+            LICENSE
```
DavidMazarro, line 55: I was thinking that we could add some more things to the release. For example, many projects include a tarball of the projects source code at the point in which the commit tag was created. I think that would be easy to add (the one created with `cabal sdist` would work I think) and could be nice to have. Maybe we could address this in some other PR though.

I'm sure there was something else I had in mind some weeks ago but I can't remember right now.

DavidMazarro:
Looks good overall, thanks for your efforts @sestrella!

.github/workflows/draft.yml
sestrella: @DavidMazarro adding the output of the `cabal sdist` command as well as uploading a tarball with the source code makes sense to me, however, I think we could add that later as you suggested taking into account this PR is already quite large :)

.github/workflows/README.md
sestrella: @DavidMazarro I really like this diagram! Thank you for the suggestion

DavidMazarro:
LGTM :shipit:

DavidMazarro (approved):

PR #201, opened 2022-07-01
Refactor hackage job using an external workflow

sestrella:


.github/workflows/release.yml
```diff
@@ -20,41 +20,7 @@ jobs:
...
-      - name: Build Haddock documentation
-        run: cabal v2-haddock --haddock-for-hackage --enable-documentation
-      - name: Upload Haddock docs to Hackage
-        run: |
-          cabal upload --publish \
-            --username "${{ secrets.HACKAGE_USERNAME }}" \
-            --password "${{ secrets.HACKAGE_PASSWORD }}" \
-            --documentation \
-            dist-newstyle/hapistrano-*-docs.tar.gz
+    uses: stackbuilders/reusable-workflows/.github/workflows/cabal-upload.yml@main
```
DavidMazarro, line 23: What does the `@main` do here?

DavidMazarro (approved):
It's nice to see this proposal moving forward!

.github/workflows/release.yml
sestrella: It is the branch name, a tag or commit hash works as well

PR #205, opened 2022-09-08
[NO ISSUE] Refactoring: Move Deploy task into library

CristhianMotoche:
### Changes
- Include restart command call inside rollback function
- Abstract deploy command into a new function

### How is this useful?
- This will help to test `deploy` easily since it has been extracted from `app/Main`
- Main becomes easier to read since the Command type constructor take us to the appropriate functions


DavidMazarro (approved):
LGTM :shipit:

PR #207, opened 2022-09-25
check if the path exists before rollback to a specific release.

### LINKED ISSUE #206: Hapistrano can fail at rollback if `ctoken` points to a release that does not exists
Hapistrano uses `ctokens` to know which deployments finished successfully. After that, they're used to rollback to a successful release. However, we're not validating if the path that is generated for the rollback exists and `ln` doesn't fail if the target directory/file doesn't exists. Therefore, we should add a validation before that symlink is created. If the file doesn't exist, we should stop the rollback and notify the user.

This could happen if at some point the `ctokens` are manually manipulated or if a release is deleted and so are the ctokens.

Example:
```
/var/project/myproject
     /ctokens/
         20220102 # Just one ctoken pointing to an non-existing release
     /releases/
         20220103
         20220104
     current -> /var/project/myproject/releases/20220104
```
Executing `hap rollback` will cause:
```
/var/project/myproject
     /ctokens/
         20220102 # Just one ctoken pointing to an non-existing release
     /releases/
         20220103
         20220104
     current -> /var/project/myproject/releases/20220102  # This symlink doesn't exist but `ln` doesn't care and doesn't fail.
```
The `current` symlink points to a directory that doesn't exists. We could face errors later because of this.

hughjfchen:
This PR fix #206

### DIFF
--- a/src/System/Hapistrano.hs
+++ b/src/System/Hapistrano.hs
@@ -12,7 +12,6 @@
 {-# LANGUAGE RecordWildCards     #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TemplateHaskell     #-}
-{-# LANGUAGE TypeApplications    #-}
 
 module System.Hapistrano
   ( runHapistrano
@@ -21,7 +20,6 @@ module System.Hapistrano
   , activateRelease
   , linkToShared
   , createHapistranoDeployState
-  , deploy
   , rollback
   , dropOldReleases
   , playScript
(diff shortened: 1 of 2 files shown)


CristhianMotoche (commented):
Thanks for the contribution @hughjfchen Although, I think there are some undesired changes in your branch. I presume your branch is outdated with the last version of hapistrano. Could you please update it?

hughjfchen:
Yeah, I'd found that just after submit the PR. The strange thing is I'd already click the `sync the fork` button on my forked repository and it says my fork is not behind the upstream and no need to update. I'm out of idea why this happens.

CristhianMotoche:
Hi @hughjfchen I assume github made something unexpected when you hit that sync button. I'd suggest to identify you changes and start from scratch:

```
git remote add sb-hap git@github.com:stackbuilders/hapistrano.git
git fetch --all
git checkout sb-hap/master -- hapistrano.cabal src/System/Hapistrano.hs
$EDITOR src/System/Hapistrano.hs # Make your changes
git add src/System/Hapistrano.hs
git commit -m "..."
git push
```

Besides that, could you please add a test for your changes? Thanks in advance!

hughjfchen:
I've updated my fork from scratch and added a test case for this PR.

src/System/Hapistrano.hs
```diff
@@ -120,10 +120,14 @@ activateRelease
...
-  exec (Ln ts rpath tpath) (Just release) -- create a symlink for the new candidate
-  exec (Mv ts tpath cpath) (Just release) -- atomically replace the symlink
+  isRpathExist <- doesDirExist rpath
+  if isRpathExist then do
+    let tpath = tempSymlinkPath deployPath
+        cpath = currentSymlinkPath deployPath
+    exec (Ln ts rpath tpath) (Just release) -- create a symlink for the new candidate
+    exec (Mv ts tpath cpath) (Just release) -- atomically replace the symlink
+  else
+    failWith 1 (Just "The path to the release you want to rollback to does not exist.") (Just release)
```
CristhianMotoche, line 130: @hughjfchen `activateRelease` is used in other places besides the `rollback` function. Let's change this description to be more general and include the path that is missing:
```suggestion
    failWith 1 (Just $ "The release path: " <> show rpath <> " cannot be activated.") (Just release)
```

CristhianMotoche (commented):
@hughjfchen It looks good to me. I just left one last comment. After that, I think we can merge this. Thanks!

CristhianMotoche (approved):

CristhianMotoche:
Thanks for the contribution @hughjfchen

PR #213, opened 2023-02-11
fix: Filter for multiple targets

CristhianMotoche:
# Description

I noticed we had a wrong filter for unique targets that was getting only the
first target. I fixed that so it can take all unique hosts.

### DIFF
--- a/spec/System/HapistranoConfigSpec.hs
+++ b/spec/System/HapistranoConfigSpec.hs
@@ -1,19 +1,22 @@
-{-# LANGUAGE CPP             #-}
-{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE CPP               #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TemplateHaskell   #-}
 
 module System.HapistranoConfigSpec
   ( spec
   ) where
 
+import qualified Data.Aeson               as A
 import           System.Hapistrano.Config (Config (..), Target (..))
-import           System.Hapistrano.Types  (Shell (..),
-                                           Source (..), TargetSystem (..))
+import           System.Hapistrano.Types  (Shell (..), Source (..),
+                                           TargetSystem (..))
 
 import qualified Data.Yaml.Config         as Yaml
 #if MIN_VERSION_base(4,15,0)
 import           Path                     (mkAbsDir, mkRelDir, mkRelFile)
 #else
-import           Path                     (mkAbsDir, mkRelDir, mkRelFile, Abs, Rel, Dir, File)
+import           Path                     (Abs, Dir, File, Rel, mkAbsDir,
+                                           mkRelDir, mkRelFile)
 #endif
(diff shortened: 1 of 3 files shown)


CristhianMotoche:
Hey @sestrella Could you please take a look at this? Also, would you mind giving me a hand with the CI? It seems an issue with the ubuntu repositories.

spec/System/HapistranoConfigSpec.hs
```diff
@@ -36,6 +39,27 @@ spec =
         Yaml.loadYamlSettings ["fixtures/git_repository_config.yaml"] [] Yaml.useEnv
           >>= (`shouldBe` defaultConfiguration)
 
+    describe "Config From JSON instances" $ do
+      context "when contains unique targets" $
+        it "parses all unique targets" $
+          let mbHosts = fmap targetHost <$> configHosts <$> A.decode content
+              content =
+                "{\"targets\":[{\"host\":\"user@app1.com\"},{\"host\":\"user@app2.com\"}],\"deploy_path\":\"/tmp\",\"local_directory\":\"/\"}"
+          in maybe
```
sestrella, line 48: @CristhianMotoche try comparing the values directly without using a `maybe` function:

```hs
let mHosts = fmap targetHost <$> configHosts <$> A.eitherDecode' content
    content = ...
in mHosts == Right ["user@app1.com", "user@app2.com"]
```

Also, I would recommend using `eitherDecode'` instead of `decode` to retrieve the original error message

CristhianMotoche: Makes sense @sestrella Thank!

CristhianMotoche: Thanks for the suggestion @sestrella Please, let me know if it looks better now. 👍

spec/System/HapistranoConfigSpec.hs
```diff
@@ -36,6 +39,27 @@ spec =
...
+    describe "Config From JSON instances" $ do
+      context "when contains unique targets" $
+        it "parses all unique targets" $
+          let eiHosts = fmap targetHost <$> configHosts <$> A.eitherDecode content
+              content =
+                "{\"targets\":[{\"host\":\"user@app1.com\"},{\"host\":\"user@app2.com\"}],\"deploy_path\":\"/tmp\",\"local_directory\":\"/\"}"
+          in case eiHosts of
+              Left msg -> fail msg
+              Right hosts ->
+                hosts `shouldBe` ["user@app1.com", "user@app2.com"]
```
sestrella, line 51: @CristhianMotoche I think this pattern match is not necessary, if you use `shouldBe` it should take care of printing the expected and got value, in the case of a `Left` it should include the error message

CristhianMotoche: Good catch @sestrella !!!

sestrella (approved):
@CristhianMotoche LGTM

PR #219, opened 2023-02-18
Add support for lead target commands

### LINKED ISSUE #218: Tasks specific for every target
Given the following scenario:
```
targets:
  - host: user@host1.com
  - host: user@host2.com

build_script: |
   build_app
   run_migrations
```

If both hosts communicate with the same database then we'll have some concurrency issues since both could run the migrations at the same time. Beyond that, there might be cases in which we want to execute tasks only in one of the hosts. Therefore, we should define a way in which we'll run target specific build scripts.

sestrella:
Closes #218

### DIFF
--- a/spec/System/HapistranoSpec.hs
+++ b/spec/System/HapistranoSpec.hs
@@ -53,37 +53,37 @@ spec = do
   describe "playScript" $ around withSandbox $ do
     context "when the target is the lead server" $ do
       it "contains the output of the lead command" $ \(deployPath, repoPath) -> do
-        output <- capture_ $ runHap $ do
+        capturedOutput <- capture_ $ runHap $ do
           release <- Hap.pushRelease $ mkTask deployPath repoPath
           Hap.playScript deployPath release Nothing LeadTarget
             [ BuildCommand (fromJust $ Hap.mkGenericCommand "echo \"hello world\"") LeadTarget
             ]
-        output `Hspec.shouldContain` "hello world"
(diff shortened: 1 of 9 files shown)


spec/System/HapistranoSpec.hs
```diff
@@ -399,7 +437,7 @@ withSandbox action =
 populateTestRepo :: Path Abs Dir -> IO ()
 populateTestRepo path =
   runHap $ do
-    justExec path "git init"
+    justExec path "git init -b master"
```
CristhianMotoche, line 440: 💯

src/System/Hapistrano/Config.hs
```diff
@@ -99,6 +104,32 @@ data Target =
     , targetSshArgs :: [String]
     } deriving (Eq, Ord, Show)
 
+data BuildCommand = BuildCommand
+  { buildCommandCommand  :: GenericCommand
+  , buildCommandOnlyLead :: ExecutionMode
```
CristhianMotoche, line 109: Hey @sestrella I think the name `buildCommandOnlyLead` can be misleading. It makes me think the value is a boolean rather than a `ExecutionMode`. What about renaming it to `buildCommandExecMode`?

CristhianMotoche (commented):
Hey @sestrella Nice test coverage! I think the PR looks good. I just left a suggestion to rename a record attribute.

CristhianMotoche (approved):
:shipit:

Issue #182, opened 2022-03-30
Replace Dockerfile base with distroless


sestrella:
Taking the [Distroless Rust](https://github.com/GoogleContainerTools/distroless/blob/main/examples/rust/Dockerfile) example as a reference, the existing `Dockerfile` could be split into different stages:

- One that compiles the project and generates a binary
- Another one that places that binary in a distroless image

The outcome would be a very tiny image. Taking into account that `hapistrano` is a deployment tool that is mostly executed by a CI pipeline, shrinking the image is beneficial since it would reduce the overhead produced by pulling the image from Docker Hub to the CI worker.

FranzGB:
> Hey @FranzGB I think you've been working on this issue lately. Any comments/questions so far?

Hey Cristhian. Yes, I am working on this still! I used a Distroless base for hapistrano I am having trouble testing if it worked.

CristhianMotoche:
Hey @sestrella. @FranzGB has been working on this in #220 but I think I faced a blocker and that is the fact that Hapistrano executes `ssh` (and maybe others, too) from the host machine. In this case, the distroless images won't have the `ssh` command and we'll get an error as the one that I [mention here](https://github.com/stackbuilders/hapistrano/pull/220#issuecomment-1569362565). Please, let us know if there is anything that we're missing in order to continue with this.

CristhianMotoche:
Hey @sestrella I was wondering if you know about a possible way to overcome the issues that we faced in @FranzGB's PR. Should we look for a different distro? Perhaps a small distribution? Let me know. Otherwise, we may not be able to solve this issue and we could close it.

CristhianMotoche:
Hey @sestrella We're using the `alpine` image which is very small ~5MB. Although, I see the hapistrano image is ~73MB, which I think was the motivation for this issue:

```
REPOSITORY                         TAG        IMAGE ID       CREATED        SIZE
alpine                             3.15       59be8124109e   5 weeks ago    5.34MB
ghcr.io/stackbuilders/hapistrano   v0.4.8.0   26c1a98bb4d2   5 months ago   73.1MB
```

The [hapistrano binary](https://github.com/stackbuilders/hapistrano/releases/tag/v0.4.8.0) is kind of big (~44MB) and maybe that's the reason the image is kind of big. So, I don't think this is a matter of the base image that we use but the size of the binary file. Would it make sense to create a new issue to address that instead of moving into a distroless base image? WDYT?

PR #220, opened 2023-04-05
Update Docker file for distroless base

FranzGB:
Updates docker file to use a light distribution base for hapistrano as requested on #182.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,14 +7,7 @@ COPY . .
 RUN cabal build --enable-executable-static && \
     cp $(cabal exec which hap) hap
 
-FROM alpine:3.15
-MAINTAINER Nicolas Vivar <nvivar@stackbuilders.com>
-RUN apk update && \
-    apk add \
-      ca-certificates \
-      git \
-      openssh-client
-RUN mkdir ~/.ssh
+FROM gcr.io/distroless/static-debian11
+LABEL maintainer="Nicolas Vivar <nvivar@stackbuilders.com>"
 COPY --from=build /usr/src/app/hap /usr/local/bin/hap
 ENTRYPOINT ["/usr/local/bin/hap"]
-CMD ["--help"]


Dockerfile
```diff
@@ -7,14 +7,7 @@ COPY . .
...
-    apk add \
-      ca-certificates \
-      git \
-      openssh-client
-RUN mkdir ~/.ssh
+FROM gcr.io/distroless/static-debian11
+LABEL maintainer="Nicolas Vivar <nvivar@stackbuilders.com>"
 COPY --from=build /usr/src/app/hap /usr/local/bin/hap
 ENTRYPOINT ["/usr/local/bin/hap"]
-CMD ["--help"]
```
CristhianMotoche, line 20: Hey @FranzGB Can we keep the `CMD ["--help"]` line?

CristhianMotoche:
Hey @FranzGB Do you need a hand with this? I think it looks fine. Is there something missing for it to be ready for review?

CristhianMotoche:
Hey @FranzGB I was finally able to test your PR in my personal computer. Unfortuantely, I got the following error when I tried to run the container with the hapistrano version in the distroless repository:

```zsh
➜  hapistrano git:(distroless-Dockerfile-base) docker images                                                                              
REPOSITORY                           TAG          IMAGE ID       CREATED          SIZE
<none>                               <none>       01b7365e6821   20 minutes ago   49.5MB
...
➜  hapistrano git:(distroless-Dockerfile-base) cd example                                            
➜  example git:(distroless-Dockerfile-base) docker run -v $PWD:/app 01b73 -c ./app/hap.yaml deploy
hap: ssh: readCreateProcessWithExitCode: posix_spawnp: does not exist (No such file or directory)
```

I guess that is expected since (I guess) the distroless image doesn't contain the `ssh` binary (and I guess it doesn't have `git` either). Therefore, I don't think this will solve #182

sestrella:
@CristhianMotoche thank you for the detailed explanation, in that case, I think it would be worth exploring SSH alternatives written in Haskell, there might be some bindings already. If we manage to decouple `hapistrano` from the SSH binary, I think it would be easier to produce a standalone binary

CristhianMotoche:
> @CristhianMotoche thank you for the detailed explanation, in that case, I think it would be worth exploring SSH alternatives written in Haskell, there might be some bindings already. If we manage to decouple `hapistrano` from the SSH binary, I think it would be easier to produce a standalone binary

Hey @sestrella Unfortunately, there are many more binary dependencies like [scp](https://github.dev/stackbuilders/hapistrano/blob/d0c08771c455ab037c3cb5958c36e68bbcda94da/src/System/Hapistrano/Core.hs#L118-L119), [git](https://github.dev/stackbuilders/hapistrano/blob/d0c08771c455ab037c3cb5958c36e68bbcda94da/src/System/Hapistrano/Commands/Internal.hs#L248-L249), and [others](https://github.dev/stackbuilders/hapistrano/blob/d0c08771c455ab037c3cb5958c36e68bbcda94da/src/System/Hapistrano/Commands/Internal.hs). It's not going to be an easy to task to replace all of those binaries.

sestrella:
@CristhianMotoche in that case, I would suggest holding a little bit on this feature, until we have a clear plan to generate a standalone binary. What do you think?

CristhianMotoche:
Sounds good to me @sestrella

CristhianMotoche:
Thanks for the contribution @FranzGB As I mentioned in #182 , let's hold on until we're sure about how to use a distroless with Hapistrano since it depends on some binaries to be present in the host (e.g. git, ssh, scp, etc.).

PR #231, opened 2023-11-20
doc: Include docs for initConfig

CristhianMotoche:


CristhianMotoche:
@cptrodolfox @DavidMazarro Can you take a look at this?

src/System/Hapistrano.hs
```diff
@@ -271,6 +269,9 @@ playScriptLocally cmds =
         }) $
   forM_ cmds $ flip execWithInheritStdout Nothing
 
+-- | Create a file with an initial config file by getting information from the
+-- user.
+
```
DavidMazarro, line 274: I'm concerned that this linebreak here might mess with Haddock properly displaying the documentation for this function, is the documentation showing up properly in the generated HTML?

src/System/Hapistrano/Config.hs
```diff
@@ -104,11 +104,13 @@ data Target =
     , targetSshArgs :: [String]
     } deriving (Eq, Ord, Show)
 
+-- | Command and execution mode for build command.
 data BuildCommand = BuildCommand
   { buildCommandCommand       :: GenericCommand
   , buildCommandExecutionMode :: ExecutionMode
   } deriving (Eq, Ord, Show)
 
+-- | Execute either on Lead or All targets.
```
DavidMazarro, line 113:
```suggestion
-- | The execution mode determines whether commands will be executed
-- on the lead target or on all targets.
```

DavidMazarro (commented):
Thanks for working on this @CristhianMotoche, looks good! I left a couple of suggestions.

src/System/Hapistrano.hs
CristhianMotoche: That's a good catch. Let's remove the empty line to keep consistency with the rest of the comments. Thanks, David.

DavidMazarro (approved):
LGTM!

PR #236, opened 2024-05-02
[225] Refactor init command

alexisbcc:


hapistrano.cabal
```diff
@@ -58,6 +58,7 @@ library
                      , filepath           >= 1.2 && < 1.5
                      , gitrev             >= 1.2 && < 1.4
                      , mtl                >= 2.0 && < 3.0
+                     , megaparsec
```
CristhianMotoche, line 61: @alexisbcc Can you please specify the boundaries of this library?

src/System/Hapistrano.hs
```diff
@@ -276,36 +278,82 @@ initConfig getLine' = do
...
+      userInput <- prompt' (parameterName <> " (default: " <> show def <> ")")
+      if null userInput then
+        pure def
+      else do
+        let parsed = M.parse (parser <* M.eof) "" userInput
+        case parsed of
+          Left err -> do
+            hPutStrLn stderr (M.errorBundlePretty  err)
+            prompt parameterName def parser
+          Right res -> pure res
```
CristhianMotoche, line 303: @alexisbcc What do you think about something like this:
```suggestion
      when (null userInput)
        pure def

      either
          (\err -> hPutStrLn stderr (M.errorBundlePretty err) >> prompt parameterName def parser)
          pure
          (M.parse (parser <* M.eof) "" userInput)
```
?

src/System/Hapistrano.hs
```diff
@@ -276,36 +278,82 @@ initConfig getLine' = do
...
+
+    promptYN = do
+      userInput <- prompt "Include restart command? y/N" 'N' yNParser
+      case toLower userInput of
+        'y' -> pure $ Just "echo 'Restart command'"
+        _ -> pure Nothing
```
CristhianMotoche, line 309:
```suggestion
      pure $ case toLower userInput of
        'y' -> Just "echo 'Restart command'"
        _ -> Nothing
```

src/System/Hapistrano.hs
```diff
@@ -276,36 +278,82 @@ initConfig getLine' = do
...
+
+    prompt' :: String -> IO String
+    prompt' title = do
+        hPutStrLn stderr title
+        getLine'
```
CristhianMotoche, line 314:
```suggestion
    prompt' title = hPutStrLn stderr title >> getLine'
```

CristhianMotoche: Why printing the title in `strerr`? What about using `putStrLn`?
```suggestion
    prompt' title = putStrLn title >> getLine'
```

src/System/Hapistrano.hs
alexisbcc: Looks better thanks

src/System/Hapistrano.hs
alexisbcc: I'm not sure about this one, I was reusing that part of the implementation :thinking:

PR #245, opened 2024-09-21
Simplification of Nix flake and removal of GHC 8

DavidMazarro:
This PR simplifies the Nix flake and GHC versions of the flake by removing GHC 8. It also:
- Bumps GHC 9 version from 9.0.2 to 9.6.6 and `transformers` dependency upper bound.
- Changes the GHC-musl image to [the one mentioned here](https://github.com/utdemir/ghc-musl/pull/42#issuecomment-1983312021), since the previous one we were using seems to be unmaintained.
- Adds Cabal, HLint and HLS to the development shell.
- Adds a `.envrc` file for use with `direnv`.
- Updates `flake.lock`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,8 +15,6 @@ jobs:
           - macos-13      # x86_64-darwin
           - macos-latest  # aarch64-darwin
           - ubuntu-latest # x86_64-linux
-        ghc:
-          - ghc966
       fail-fast: true
     runs-on: ${{ matrix.os }}
     timeout-minutes: 30
@@ -27,9 +25,9 @@ jobs:
           extra-conf: accept-flake-config = true
       - uses: DeterminateSystems/magic-nix-cache-action@v7
       - name: Compile code ${{ matrix.ghc }}
-        run: nix build .#test-${{ matrix.ghc }}
+        run: nix build .#test
       - name: Run tests ${{ matrix.ghc }}
-        run: nix run .#test-${{ matrix.ghc }}
+        run: nix run .#test
 
   docker:
     uses: ./.github/workflows/reusable-docker.yml
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,8 +16,7 @@ jobs:
           - macos-latest  # aarch64-darwin
           - ubuntu-latest # x86_64-linux
         ghc:
-          - ghc8107
-          - ghc902
+          - ghc96
       fail-fast: true
     runs-on: ${{ matrix.os }}
     timeout-minutes: 30
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
           - macos-latest  # aarch64-darwin
           - ubuntu-latest # x86_64-linux
         ghc:
-          - ghc96
+          - ghc966
       fail-fast: true
     runs-on: ${{ matrix.os }}
     timeout-minutes: 30
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM utdemir/ghc-musl:v24-ghc96 AS build
+FROM quay.io/benz0li/ghc-musl:9.6.6 AS build
 WORKDIR /usr/src/app
 COPY hapistrano.cabal .
 RUN cabal update && \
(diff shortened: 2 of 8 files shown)


Dockerfile
```diff
@@ -8,7 +8,7 @@ RUN cabal build --enable-executable-static && \
     cp $(cabal exec which hap) hap
 
 FROM alpine:3.15
-MAINTAINER Nicolas Vivar <nvivar@stackbuilders.com>
+MAINTAINER Cristhian Motoche <cmotoche@stackbuilders.com>
```
CristhianMotoche, line 11: 😊

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM utdemir/ghc-musl:v24-ghc902 AS build
+FROM quay.io/benz0li/ghc-musl:9.6.6 AS build
```
sestrella, line 1: @DavidMazarro was the previous image deprecated?

flake.nix
```diff
@@ -27,51 +27,34 @@
...
-                  src = ./.;
+                # This is used by `nix develop .` to open a shell for use with
+                # `cabal`, `hlint` and `haskell-language-server`
+                shell.tools = {
+                  cabal = {};
+                  hlint = {};
+                  haskell-language-server = {};
                 };
-                compiler-nix-name = "ghc902";
+                compiler-nix-name = "ghc966";
```
sestrella, line 42: @DavidMazarro If we plan to use only one version of GHC to compile the project moving forward, I believe we could refactor this further:

```nix
{
  ...
  outputs = { self, nixpkgs, flake-utils, haskellNix }:
    flake-utils.lib.eachSystem [ ... ] (system:
    let
      overlays = [ haskellNix.overlay
        (final: prev: {
          # This overlay adds our project to pkgs
          hapistrano =
            final.haskell-nix.project' {
              src = ./.;
              compiler-nix-name = "ghc966";
              # This is used by `nix develop .` to open a shell for use with
              # `cabal`, `hlint` and `haskell-language-server`
              shell.tools = {
                cabal = {};
                hlint = {};
                haskell-language-server = {};
              };
              # Non-Haskell shell tools go here
              shell.buildInputs = with pkgs; [
                nixpkgs-fmt
              ];
              # This adds `js-unknown-ghcjs-cabal` to the shell.
              # shell.crossPlatforms = p: [p.ghcjs];
            };
        })
      ];
      pkgs = import nixpkgs { inherit system overlays; inherit (haskellNix) config; };
      flake = pkgs.hapistrano.flake {
        # This adds support for `nix build .#js-unknown-ghcjs:hello:exe:hello`
        # crossPlatforms = p: [p.ghcjs];
      };
    in flake // rec {
      # Built by `nix build .`
      packages= {
        default = flake.packages."hapistrano:exe:hap";
        test = flake.packages."hapistrano:test:test".overrideAttrs(...);
      };
      apps = {
        type = "app";
        program = "${packages.test}/bin/test";
      }
    });
}
```

Source: https://input-output-hk.github.io/haskell.nix/tutorials/getting-started-flakes.html#scaffolding

hapistrano.cabal
```diff
@@ -28,7 +28,7 @@ category:            System
 homepage:            https://github.com/stackbuilders/hapistrano
 bug-reports:         https://github.com/stackbuilders/hapistrano/issues
 build-type:          Simple
-tested-with:         GHC==8.10.17, GHC==9.0.2
+tested-with:         GHC==9.6.5
```
sestrella, line 31:
```suggestion
tested-with:         GHC==9.6.6
```

sestrella: @DavidMazarro I believe we should also bump the version of `base` at some point; however, I believe it would be best to do so in a separate PR because it would most likely require a major release.

.github/workflows/build.yml
```diff
@@ -16,8 +16,7 @@ jobs:
           - macos-latest  # aarch64-darwin
           - ubuntu-latest # x86_64-linux
         ghc:
-          - ghc8107
-          - ghc902
+          - ghc966
```
sestrella, line 19: @DavidMazarro If we are only going to use one version of GHC in the future, I believe we can remove the GHC version from the package name:

```diff
- nix run .#test-ghc966
+ nix run .#test
```

flake.nix
DavidMazarro: Good point. Yeah, I was following that same documentation when I added the shell tools 😄

Dockerfile
DavidMazarro: Not deprecated per se, but it is unmaintained. See [this comment](https://github.com/utdemir/ghc-musl/pull/42#issuecomment-1983312021) where I learned about the alternative.

sestrella:
@DavidMazarro nice job!

PR #249, opened 2024-11-12
Migrate of Nix flake to devenv and addition of an HLint pre-commit hook

### LINKED ISSUE #238: Include hlint as part of the pre-commit and CI process
It would be a good idea to follow hlint suggestions before making a commit or pushing changes to the `main` branch.

DavidMazarro:
Closes #238.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -34,12 +34,18 @@
           inherit inputs pkgs;
           modules = [
             {
+              # Need to set GHC_NO_UNICODE as mentioned here:
+              # https://gitlab.haskell.org/ghc/ghc/-/commit/74132c2b0992849f83ef87c8a56ac3975738e767#af1d9e2647379d23697be69f8b7fc568ed0294c5_1825_1825
+              env.LC_ALL = "en_US.UTF-8";
+              env.GHC_CHARENC = "GHC_NO_UNICODE";
+
               languages.haskell.enable = true;
               languages.haskell.package = pkgs.haskell.compiler.ghc96;
 
               # https://devenv.sh/reference/options/
               packages = with pkgs; [
                 pre-commit
+                zsh
               ];
 
               enterShell = ''
@@ -47,6 +53,7 @@
               '';
 
               enterTest = ''
+                cabal update
                 cabal test
               '';
             }
--- a/flake.nix
+++ b/flake.nix
@@ -36,10 +36,13 @@
                 hlint = {};
                 haskell-language-server = {};
               };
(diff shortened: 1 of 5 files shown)


flake.nix
```diff
@@ -1,68 +1,64 @@
...
+              # https://gitlab.haskell.org/ghc/ghc/-/commit/74132c2b0992849f83ef87c8a56ac3975738e767#af1d9e2647379d23697be69f8b7fc568ed0294c5_1825_1825
+              env.LC_ALL = "en_US.UTF-8";
+              env.GHC_CHARENC = "GHC_NO_UNICODE";
+
+              languages.haskell.enable = true;
+              languages.haskell.package = pkgs.haskell.compiler.ghc96;
+
+              # https://devenv.sh/reference/options/
+              packages = with pkgs; [
+                pre-commit
```
oscar-izval, line 47: I think you can simplify the setup a bit by defining the pre-commit hooks with the implementation that devenv exposes by default. You can find more details [here](git-hooks.hooks.hlint). That way you don't have to track the config file in git and you can get rid of the entershell command

flake.nix
```diff
@@ -1,68 +1,64 @@
...
-                ]}
-            '';
-          });
+  outputs = {
+    self,
+    nixpkgs,
+    devenv,
+    systems,
+    ...
+  } @ inputs: let
```
sestrella, line 20: @DavidMazarro I'm wondering if the complexity of this file comes from the fact that we are using it for two different purposes. On one hand, it appears that we want a portable development environment, and on the other, we want to provide a Nix flake version of this project that can be properly referenced in other Nix projects. If that's the case, I'd recommend creating a standalone `devenv.nix` file for local development and leaving the current `flake.nix` with only the `packages` declaration; this should simplify some maintenance tasks as `devenv` is a higher-level abstraction over Nix flakes, making it more approachable for new comers.

flake.nix
```diff
@@ -1,68 +1,64 @@
...
-        flake = pkgs.hapistrano.flake { };
-      in rec {
-        apps = {
-          test = {
-            type = "app";
-            program = "${packages.test}/bin/test";
-          };
-        };
-        packages = {
-          default = flake.packages."hapistrano:exe:hap";
```
sestrella, line 53: @DavidMazarro If we intend to remove the 'default' package, we should consider doing so as a breaking change, as other Nix flakes may already be referencing this project.

DavidMazarro:
Closing in favor of #251. I will be adding the HLint pre-commit hook on a separate PR.

PR #251, opened 2025-03-25
chore: Replace `haskell.nix` with `stacklock2nix`

sestrella:
This PR aims to simplify the current Nix setup for building the project by replacing `haskell.nix` with `stacklock2nix`, which has a simple interface and generates Nix derivations using the existing `stack.yaml.lock`. Additionally, it configures 'devenv' to provide Nix contributors with a development environment shell that includes all of the tooling needed to build and test the project.

### Q&A

@fm7-1, @CristhianMotoche, @DavidMazarro I'm just trying to provide some background for the thought process behind some of the changes included in this PR; I'd be happy to discuss this with you if you have any concerns.

**Why did the stack resolver get upgraded?**

GHC 8.10.4 does not have great support for Apple Silicon, but given that Hapistrano is intended to be used as a tool rather than a library, upgrading the compiler to a more recent version should not be a major issue.


.github/workflows/build.yml
```diff
@@ -9,31 +9,76 @@ on:
...
   group: build-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
-  haskell:
+  build:
+    strategy:
+      matrix:
+        runner:
+          - macos-latest
```
fm7-1, line 22: Hi, @sestrella! I'm wondering if we should also build for Intel-based macOS? 🤔 Probably not, but wanted to mention it

.github/workflows/build.yml
```diff
@@ -9,31 +9,76 @@ on:
...
+    runs-on: ${{ matrix.runner }}
     steps:
-      - uses: actions/checkout@v4
-      - uses: DeterminateSystems/nix-installer-action@v12
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Install Nix
+        uses: cachix/install-nix-action@v31
+      - name: Setup Cachix
+        uses: cachix/cachix-action@v16
```
fm7-1, line 76: 💯

.github/workflows/build.yml
fm7-1: Or maybe we do, I see `x86_64-darwin` is part of the supported systems in the flake file

DavidMazarro: Considering that Intel-based macOS will be fully out of support in a couple of years, I think it's best we stick to Apple Silicon for macOS runners from now on

.github/workflows/build.yml
```diff
@@ -9,31 +9,78 @@ on:
...
+      - name: Setup cache for `~/.stack`
+        uses: actions/cache@v4
+        with:
+          path: ~/.stack
+          key: ${{ runner.os }}-stack-global-${{ hashFiles('stack.yaml') }}-${{ hashFiles('hapistrano.cabal') }}
+          restore-keys: |
+            ${{ runner.os }}-stack-global-${{ hashFiles('stack.yaml') }}-
+            ${{ runner.os }}-stack-global-
+      - name: Install dependencies
+        run: devenv shell -- stack build --only-dependencies --test
```
DavidMazarro, line 48: Do we need `--test` for dependencies?

.github/workflows/build.yml
```diff
@@ -9,31 +9,78 @@ on:
...
       matrix:
-        os:
-          - macos-13      # x86_64-darwin
-          - macos-latest  # aarch64-darwin
-          - ubuntu-latest # x86_64-linux
-      fail-fast: true
-    runs-on: ${{ matrix.os }}
-    timeout-minutes: 30
+        runner:
+          - macos-13         # x86_64-darwin
```
DavidMazarro, line 67: I don't think we need to do this yet, but I would eventually consider phasing out Intel based macOS builds: the `macos-12` runner disappeared last year, and it's only a matter of time before this one is unsupported too (and it's the last Intel based macOS runner available for GH Actions 😬)

flake.nix
```diff
@@ -1,68 +1,41 @@
...
+      self,
+      nixpkgs,
+      stacklock2nix,
+      ...
+    }:
+    let
+      supportedSystems = [
+        "aarch64-darwin"
+        "aarch64-linux"
+        "x86_64-darwin"
```
DavidMazarro, line 18: Similar to the other comment I posted, I think we eventually want to stop officially supporting x86 macOS. If we keep it as a supported system, we should warn that users are "on their own".

DavidMazarro (commented):
Great work Sebas! So much simpler :) I left a couple of comments

.github/workflows/build.yml
sestrella: @DavidMazarro sounds good; I guess we could `stack build` against the most recent Ubuntu and MacOS runners. Regarding the Nix build, I was able to build against all of the different Linux and Darwin variants (`aarch64` and `x86_64`); however, we reached the maximum storage limit on Cachix, so it began removing some things, resulting in long build times on the CI; thus, I believe keeping `ubuntu-latest` and `macos-latest` should suffice for the time being.

image

https://www.cachix.org/pricing

.github/workflows/build.yml
sestrella: @DavidMazarro Yes! We need it to build the test dependencies as well.

flake.nix
sestrella: Please see the comment below

hapistrano.cabal
```diff
@@ -52,13 +52,13 @@ library
                      , System.Hapistrano.Commands.Internal
                      , System.Hapistrano.Maintenance
   build-depends:       aeson              >= 2.0 && < 3.0
-                     , ansi-terminal      >= 0.9 && < 0.12
+                     , ansi-terminal      >= 0.9 && < 2.0
```
sestrella, line 55: I am keeping the original lower bound for all the dependencies where I bumped the upper bound to avoid introducing some breaking changes.

DavidMazarro (approved):
LGTM!

